PR #356, opened 2014-05-07
RHEL has a separate services for IPv4 and IPv6 iptables

tmclaugh:


jonnytpuppet:
@csschwe @igalic why do you feel this feature should be optional? To my understanding, there is no "ipv4 only" bool for using this module with Debian?

Also, not starting the ip6table service on RHEL won't actually stop ip6tables from working; to my understanding, it just stops the rules being saved at shutdown and applied at boot from the /etc/ip6tables file...I think

Cheers!

csschwe:
I fully disable and stop ip6tables service and any IP6 kernel features on my RHEL servers.   Having the services tied together means I will have to customize this module in my environment to ignore/disable ip6tables.

jonnytpuppet:
@csschwe I do not believe that the iptables/ip6tables service actually disables the linux firewall per-se. To my understanding, it simply disables the saving of rules upon shutdown, and the reloading of rules (and other default configuration) upon boot. This is similar to debian's iptables-persistant service.

I am of the opinion that we should keep the module persistent across all oses, and just enable both iptables and ip6tables service together on RHEL, just like we enable debian's iptables-persistent service (which does both iptables and ip6tables).

If one wishes to disable ipv6, turning off ip6tables service on RHEL won't do anything. You need to change your kernel sysctl options or grub boot options, which is outside the scope of this module.

@cmurphy @mhaskel @hunner Do you guys have an opinion on this?

csschwe:
You are mostly correct.  The only issue is that the service script in RHEL includes a stanza that looks for the presence of " install ipv6 /bin/true" in /etc/modprobe.d/*.   If found the service script will return a 150 return code.

RedHat instructions to fully disable ipv6 include the following:
To prevent the module from loading, modify /etc/modprobe.d/ipv6.conf with the following only:
install ipv6 /bin/true

If you don't have a Red Hat login you will probably not be able to see this link.
https://access.redhat.com/solutions/8709

mhaskel:
@tmclaugh @csschwe @jonnytpuppet This should be possible after #436 since that added the `$service_name` parameter. I think it's probably best to leave the current defaults, and not include the ip6tables service on el unless a user specifically wants it.

Since I think the original intent of the PR is now possible, I'm going to close this request for now. If it's found that more or different functionality is needed, please reopen or submit a new pull request.

Thanks for the work on this!

PR #405, opened 2014-09-03
EL7 Fixes for CentOS

cdenneen:


lib/puppet/util/firewall.rb
```diff
@@ -169,7 +169,7 @@ def persist_iptables(proto)
     end
 
     # RHEL 7 and newer also use systemd to persist iptable rules
-    if os_key == 'RedHat' && Facter.value(:operatingsystem) == 'RedHat' && Facter.value(:operatingsystemrelease).to_i >= 7
+    if os_key == 'RedHat' && ['RedHat','CentOS','Scientific','SL','SLC','Ascendos','CloudLinux','PSBM','OracleLinux','OVS','OEL','Amazon','XenServer'].include?(Facter.value(:operatingsystem)) && Facter.value(:operatingsystemrelease).to_i >= 7
```
mlehner616: Many of the OSs listed use different versioning from RedHat and CentOS. Probably the most common among those listed is Amazon which uses YYYY.MM versioning. So, for Amazon this condition will always be true which could cause unexpected behavior.

PR #440, opened 2014-12-12
(MODULES-1592) Allow src_type and dst_type prefixed with '! ' to validate

jstangroome:


jonnytpuppet:
Hi There

Thanks for your submission!

Unfortunately, your implementation will not work on older versions of iptables that do not support multiple invocations of the same "-m"; in your case, "-m addrtype" would be called twice if a user specified both src-type and dst-type.

I came across this same issue when I was implementing the following physdev properties for this module: physdev-in, physdev-out and physdev-is-bridged - all of these use the "-m physdev" iptables argument.

Please look at the following lines in the iptables provider to see how i've implemented this:

https://github.com/puppetlabs/puppetlabs-firewall/blob/master/lib/puppet/provider/firewall/iptables.rb#L255

https://github.com/puppetlabs/puppetlabs-firewall/blob/master/lib/puppet/provider/firewall/iptables.rb#L461

https://github.com/puppetlabs/puppetlabs-firewall/blob/master/lib/puppet/provider/firewall/iptables.rb#L108

In the meantime I will close this PR as the implementation is unsuitable for the reasons listed above. I encourage you to re-implement your solution and open a new PR using a similar design to how I physdev, in order to maintain support for older iptables versions.

Also, please remember to add acceptance tests in the spec/acceptance directory.

Thanks!

jonnytpuppet:
Reopening pull request as issue identified about is not due to this PR

jonnytpuppet:
@jstangroome it would be appreciated if you could write some acceptance tests for this in the spec/acceptance directory

Thanks!

jonnytpuppet:
ping @jstangroome is there any update on this? Thanks

jstangroome:
Sorry @jonnytpuppet I'm not working in this space at the moment and cannot be sure when I'll be able to come back to it to write some tests

jonnytpuppet:
Hi There

Thanks for your contribution!

I'm closing this due to inactivity. Feel free to re-open another PR with the above comments addressed.

Thanks

PR #447, opened 2014-12-20
Use rspec-puppet-facts

mcanevet:


jonnytpuppet:
Hi There

Thanks for your submission!

Unfortunately we can't merge this as the tests are failing. Also, you've added support for ArchLinux which requires separate discussion, and would be part of a separate PR.

Please re-open another pull request if you feel you have made changes that are suitable.

Cheers

PR #449, opened 2014-12-25
fixed pkg and service name on Ubuntu 14.10

pcheliniy:
Fix an issue with init-script name on Ubuntu 14.10 (was renamed to netfilter-persistent)

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -33,6 +33,15 @@ class firewall::params {
         $package_name = 'iptables-persistent'
       }
     }
+    'Ubuntu' : {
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -33,6 +33,15 @@
         $package_name = 'iptables-persistent'
       }
     }
+    'Ubuntu' : {
```
mlehner616: The cleanest way to do this is to move this logic into the Debian case and make it look similar to the RedHat case (a nested case for operatingsystem).

mhaskel:
Fixed in https://github.com/puppetlabs/puppetlabs-firewall/pull/476

PR #466, opened 2015-01-09
Lookup username from uuid

bodepd:
When using the uid feature of the firewall module,
it did not work with string based usernames as
documented.

The uid propery always synchronized with a message of
<number> does not match <username>.

This code overrides the uid getter method to perform
a check of both the data from the property hash as well
as using that data (assuming it is a uid) to resolve the
username.

While this patch is pretty simple, I have only tested it
on Ubuntu 14.04. I am not sure if it could be problematic
with other versions.

I have not yet written tests b/c I wanted to submit
my proposed fix for discussion while I get those
written.

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -693,6 +693,11 @@ Puppet::Type.newtype(:firewall) do
       only, as iptables does not accept multiple uid in a single
       statement.
     EOS
+    def insync?(is)
+      require 'etc'
+      return is.to_s == @should.first.to_s || Etc.getpwuid(Integer(is)).name == @should.first.to_s
+    end
+
   end
 
   newproperty(:gid, :required_features => :owner) do


jonnytpuppet:
I'm using the following manifest to test this PR:

```
firewall { '300 only allow root user access':
  proto       => 'all',
  chain       => 'OUTPUT',
  uid         => '! root',
  destination => '169.254.169.254',
  action      => 'drop',
}
```
How after re-applying the same manifest, the following error is thrown:
```
Error: invalid value for Integer(): "! 0"
Error: /Stage[main]/Main/Firewall[300 only allow root user access]/uid: change from ! 0 to ! root failed: invalid value for Integer(): "! 0"
```

bodepd:
@jonnytpuppet I definitely did not think to test the negation case, I'll pick it back up and take a stab when I get a second. Thanks!

bodepd:
closing as it's been replaced by #474

PR #467, opened 2015-01-11
Add missing :mask on ip6tables provider

jonathanio:
The :mask option was missing in the ip6tables provider which prevented it from correctly working with the recent module with the following error:

Could not evaluate: Parser error: rttl was meant to be a boolean but received value: "--mask."

### DIFF
--- a/lib/puppet/provider/firewall/ip6tables.rb
+++ b/lib/puppet/provider/firewall/ip6tables.rb
@@ -26,6 +26,7 @@ Puppet::Type.type(:firewall).provide :ip6tables, :parent => :iptables, :source =
   has_feature :iprange
   has_feature :ipsec_dir
   has_feature :ipsec_policy
+  has_feature :mask
 
   optional_commands({
     :ip6tables      => 'ip6tables',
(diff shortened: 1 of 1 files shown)


mhaskel:
Thanks for the contribution @jonathanio.

This looks good, but should have doc and test updates. For docs, the [Providers](https://github.com/puppetlabs/puppetlabs-firewall#providers) section of the README should be updated to reflect that ip6tables has the mask feature. For tests, something should be added in `firewall_spec.rb` similar to https://github.com/puppetlabs/puppetlabs-firewall/blob/master/spec/acceptance/firewall_spec.rb#L1424-L1450

Let me know if you need a hand getting this work finished up, I've been working on syncing up the iptables and ip6tables providers.

mhaskel:
@jonathanio thanks for your contribution! we wanted to get the ticket I was working on closed out so I updated the tests/docs myself in #469.

jonathanio:
@mhaskel No, thank you as well. TBH, I wasn't going to be able to look at it until next week so thank you for your help too!

PR #471, opened 2015-01-15
Fixing regressions for Amazon Linux since RH7 support was added

mlehner616:
By adding support for RH7, Amazon was getting incorrectly grouped with those newer operating systems. Amazon Linux is still using the RH6 way to do things making the module unusable on all Amazon systems. I'm adding exceptions for Amazon as the conditions for RH7 were incorrectly catching Amazon Linux because the operatingsystemrelease YYYY.MM will always evaluate to `true`

### DIFF
--- a/lib/puppet/util/firewall.rb
+++ b/lib/puppet/util/firewall.rb
@@ -169,7 +169,7 @@ module Puppet::Util::Firewall
     end
 
     # RHEL 7 and newer also use systemd to persist iptable rules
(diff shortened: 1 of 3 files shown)


mlehner616:
From looking at the code, this issue could be affecting other distributions as well.

mlehner616:
@puppetcla to save time and get this thing merged, CLA has been signed.

mhaskel:
@mlehner616 Can you squash this down to a single commit?

Otherwise looks good, and just FYI the CLA isn't required for module contributions

mlehner616:
@mhaskel Not a problem, I'll squash it.

ok, figured I'd get it signed just in case, I saw it was holding up another PR (could have been another repo)

mlehner616:
@mhaskel Should be all set just waiting to hear from Travis

mhaskel:
Great, thanks @mlehner616

PR #474, opened 2015-01-21
Uid negation fix

jonnytpuppet:
Fix for MODULES-1688.

Re-applying a manifest with an unchanged UID will now not re-apply the rule unnecessarily. Works properly for negated and non-negated cases.


cyberious:
@bodepd thoughts?  You cool with us closing yours and merging this with yours as well?

jonnytpuppet:
Fixes #466

bodepd:
no problem

PR #475, opened 2015-01-21
MODULES-753 further fixes

jonnytpuppet:
Fixes user adding and removing using either UID or string username.
Tested with both negated and non-negated values and works on all
supported operating systems.

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -737,20 +737,37 @@ Puppet::Type.newtype(:firewall) do
       # between string usernames and UIDs (integers). We also need to ignore
       # spaces as they are irrelevant with respect to rule sync.
 
+      # Remove whitespace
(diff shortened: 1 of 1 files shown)


hunner:
"Tested with" <-- do you want to commit the tests too?

jonnytpuppet:
There are no new tests for this PR. The tests were committed with #474

lib/puppet/type/firewall.rb
```diff
@@ -737,20 +737,37 @@ def insync?(is)
...
-        negate = '!'
-      else
-        lookup_id = is
-        negate = ''
+          is = is.gsub(/^!/,'')
+          is_negate = '!'
+      end
+      if should.start_with?('!')
+          should = should.gsub(/^!/,'')
+          should_negate = '!'
```
mhaskel: 2-space soft tabs should be used here

PR #477, opened 2015-01-23
Modules 633

mhaskel:
If the packages for ip6tables aren't installed, the provider will throw an exception. Fix the provider to check to make sure the variable exists first, and update params.pp to include the required package for EL6.

### DIFF
--- a/lib/puppet/provider/firewall/ip6tables.rb
+++ b/lib/puppet/provider/firewall/ip6tables.rb
@@ -45,7 +45,8 @@ Puppet::Type.type(:firewall).provide :ip6tables, :parent => :iptables, :source =
 
 
   def initialize(*args)
-    if Facter.fact('ip6tables_version').value.match /1\.3\.\d/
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/firewall/ip6tables.rb
```diff
@@ -45,7 +45,8 @@
 
 
   def initialize(*args)
-    if Facter.fact('ip6tables_version').value.match /1\.3\.\d/
+    ip6tables_version = Facter.fact('ip6tables_version').value
```
elyscape: This can be replaced with:
```ruby
ip6tables_version = Facter.value('ip6tables_version')
```
Or, possibly superior:
```ruby
ip6tables_version = Facter.value(:ip6tables_version)
```

cmurphy:
Could you separate these into different pull requests or add a description for how these commits are related? I can't really tell what this pull request is about without looking at the ticket.

PR #500, opened 2015-03-11
(MAINT) Adds an acceptance test for RHEL 7

petems:
Lets us check the logic around making sure RHEL 7 changing service is working


spec/acceptance/rhel7_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper_acceptance'
+
+describe 'firewall on RHEL7', :unless => (UNSUPPORTED_PLATFORMS.include?(fact('osfamily')) || !is_rhel7( fact('osfamily'), fact('osfamily'), fact('operatingsystemrelease')) ) do
```
elyscape: Shouldn't this read:
```ruby
!is_rhel7( fact('osfamily'), fact('operatingsystem'), fact('operatingsystemrelease'))
```
Note the second argument.

petems: Good point...

petems: Fixed

PR #501, opened 2015-03-17
Hiera integration for firewall rules

discreet:
I put in a class called custom_firewall.pp which allows you to create and remove your firewall rules using Hiera.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -19,6 +19,10 @@ class firewall::params {
           }
           $service_name = 'iptables'
         }
+        'Gentoo': {
+          $service_name = ['iptables','ip6tables']
+          $package_name = 'net-firewall/iptables'
+         }
         default: {
           if versioncmp($::operatingsystemrelease, '7.0') >= 0 {
             $package_name = 'iptables-services'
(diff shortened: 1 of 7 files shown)


daenney:
It is currently missing any form of documentation and tests, it's also causing the current tests to fail.

Is there any particular reason this wasn't added on the main class as a `$rules` parameter which is bound to undef or an empty array by default instead?

elyscape:
Here's a better question: why is a personal `my_fw` implementation being submitted as a pull request to the module itself?

manifests/params.pp
```diff
@@ -18,6 +19,10 @@
           }
           $service_name = 'iptables'
         }
+        'Gentoo': {
```
elyscape: Why is Gentoo moved into the RedHat osfamily?  It's not a RedHat variant.

jonnytpuppet:
I'm closing this as this has far too many features to be considered for a single pull request. You also need to add new tests for the new features added, and all current tests must still pass.

Also, commit should be squashed

Feel free to open a new pull request with updated commits.

PR #504, opened 2015-03-25
Fix os-family for ArchLinux.

pospim19:
Starting & enabling service(s) didn't work, because ArchLinux has own os-family.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -6,10 +6,6 @@ class firewall::params {
           $service_name = 'iptables'
           $package_name = undef
         }
-        'Archlinux': {
-          $service_name = ['iptables','ip6tables']
-          $package_name = undef
-        }
         'Fedora': {
           if versioncmp($::operatingsystemrelease, '15') >= 0 {
             $package_name = 'iptables-services'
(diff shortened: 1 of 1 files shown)


hunner:
Only on newer facters... Do you know what version this changed? Perhaps we should retain both for backwards-compatibility?

pospim19:
I don't now what version this behaviour changed, probably before facter 2.4.1.
ArchLinux has rolling updates - when it's installed or updated it gets fresh software, so I don't think we need both for backwards-compatibility.

elyscape:
>Only on newer facters... Do you know what version this changed?

Are you sure?  Because Arch is not and has never been based on EL.

hunner:
Sorry, what I was thinking of is that on facter < 1.7.0 Archlinux had osfamily Linux, and 1.7.0 created an Archlinux osfamily of its own.

But yeah, doesn't matter anyway since this was under the RedHat osfamily.

manifests/params.pp
```diff
@@ -56,6 +52,10 @@
         }
       }
     }
+    'Archlinux': {
+      $service_name = ['iptables','ip6tables']
+      $package_name = undef
+     }
```
hunner: Off-by-one indentation here

jonnytpuppet:
Hi There

Thanks for your contribution!

I'm closing this due to inactivity. Please feel free to open a new PR which pass the travisci unit tests.

Thanks!

PR #488, opened 2015-02-08
README: Added example of firewallchain for drop policy on input.

jpds:

PR #506, opened 2015-03-29
README: Added example of firewallchain for drop policy on input.

jpds:
Late followup from #488.


tphoney:
Thanks for the examples. It needs the commits squashed.

PR #512, opened 2015-04-08
Enable firewall rule persistence for Red Hat-derived OSes

alexjurkiewicz:
Currently on RHEL (v6), firewall config is not saved on service stop/restart. Use of this module causes an awkward period on system boot where the firewall is empty until Puppet runs. You get the same problem when restarting iptables for whatever reason. There are two options in `/etc/sysconfig/iptables-config` which help here:

    # Save current firewall rules on stop.
    #   Value: yes|no,  default: no
    # Saves all firewall rules to /etc/sysconfig/iptables if firewall gets stopped
    # (e.g. on system shutdown).
    IPTABLES_SAVE_ON_STOP="yes"
    
    # Save current firewall rules on restart.
    #   Value: yes|no,  default: no
    # Saves all firewall rules to /etc/sysconfig/iptables if firewall gets
    # restarted.
    IPTABLES_SAVE_ON_RESTART="yes"

This module manages `/etc/sysconfig/iptables-config` in order to change these values to "yes".

To me, this change makes the module behave like I expect, but I'm not sure if you agree, or what the current behaviour is on other OSes.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -345,6 +345,7 @@ Performs the basic setup tasks required for using the firewall resources.
 At the moment this takes care of:
 
 * iptables-persistent package installation
+* On Red Hat-derived systems, managing /etc/sysconfig/iptables-config
 
 Include the `firewall` class for nodes that need to use the resources in this module:
 
--- a/files/iptables-config
+++ b/files/iptables-config
@@ -1,0 +1,56 @@
+# This file is managed by Puppet
+
+# Load additional iptables modules (nat helpers)
+#   Default: -none-
+# Space separated list of nat helpers (e.g. 'ip_nat_ftp ip_nat_irc'), which
+# are loaded after the firewall rules are applied. Options for the helpers are
+# stored in /etc/modprobe.conf.
+IPTABLES_MODULES=""
+
+# Unload modules on restart and stop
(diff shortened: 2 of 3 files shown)


alexjurkiewicz:
https://tickets.puppetlabs.com/browse/MODULES-1922

mhaskel:
Hi @alexjurkiewicz,

Thanks for the contribution! I think I'd rather have this be configurable (whether or not the file is managed, and then have the file set up as a template so that the values can be changed).

This also needs test updates to make sure it's actually working.

elyscape:
Firewall rule persistence [is in there already](https://github.com/puppetlabs/puppetlabs-firewall/blob/master/lib/puppet/util/firewall.rb#L144-L225), performed when new or modified firewall resources are applied.

mhaskel:
@elyscape good point, thanks.

@alexjurkiewicz if this isn't working for you I'd suspect there's something wrong with the persist functionality, and it would probably be better to update that rather than to add extra file management.

jonnytpuppet:
Closing due to concerns rasied by @mhaskel

Persistence is already there for puppet managed rules. If this is not working for you, please raise a bug using our JIRA.

Thanks!

PR #517, opened 2015-04-24
(MODULES-1976) Revise rule name validation for ruby 1.9

karmix:
Note: This test will only fail for ruby 1.9 and above.  I was unable to find a way to reproduce the character encoding functionality introduced in 1.9 on previous versions of ruby in order to construct a test that is useful in those environments.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -188,7 +188,7 @@ All rules employ a numbering system in the resource's title that is used for ord
 
 You can place default rules in either `my_fw::pre` or `my_fw::post`, depending on when you would like them to run. Rules placed in the `pre` class will run first, and rules in the `post` class, last.
 
-In iptables, the title of the rule is stored using the comment feature of the underlying firewall subsystem. Values must match '/^\d+[[:alpha:][:digit:][:punct:][:space:]]+$/'.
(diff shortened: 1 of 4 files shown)


bmjen:
Thanks for the contribution @karmix! The parser in this module isn't something we like to modify too frequently. Do you have a specific reason for wanting special characters in the rule name?

jonnytpuppet:
Hi There

Is this related to #513 at all?

Thanks

karmix:
No.  These are different issues.  I encountered them separately.

Thanks

jonnytpuppet:
Hi Karmix

I'm a bit confused about your statement "This test will only fail for ruby 1.9 and above". Are you saying that your changes wont work on Ruby >=1.9, or that your changes have fixed an issue that would have occured on Ruby >=1.9 otherwise?

Thanks

karmix:
This patch addresses a difference in the way unicode and non-unicode strings match regular expressions in Ruby >= 1.9.  What I was trying to say was that the test I added to spec/fixtures/iptables/conversion_hash.rb will never fail on Ruby < 1.9, where unicode support is not present.

PR #519, opened 2015-04-30
(MODULES-2000) netfilter-persistent systems need iptables-persistent pkg

gertvdijk:
With the changes in #403 and #436, systems Debian >=8 or Ubuntu >= 14.10 now don't have a dependency on `iptables-persistent` package offering the `netfilter-persistent` plugins required for persistence of iptables/iptables6 rules. Without these plugins, the `netfilter-persistent` *service* silently perfoms no action. The `iptables-persistent` *package* depends on the `netfilter-persistent` *package* and will therefore provide the right *service*.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -29,30 +29,26 @@ class firewall::params {
       }
     }
     'Debian': {
+      $package_name = 'iptables-persistent'
       case $::operatingsystem {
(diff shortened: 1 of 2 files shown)


jonnytpuppet:
Closing this due to #523

PR #541, opened 2015-06-14
(MODULES-2118) Fix severe installation hang on jessie.

fsdef:
On jessie iptables-persistent installation/configuration pops up debconf to ask if the current firewall setup should be saved which leads to puppet hanging and 99% of the time half configured firewall and ssh lockout.

This patch sets the value to false before the install so that debconf doesn't pop up.

### DIFF
--- a/manifests/linux/debian.pp
+++ b/manifests/linux/debian.pp
@@ -21,8 +21,14 @@ class firewall::linux::debian (
 ) inherits ::firewall::params {
 
   if $package_name {
+    #Fixes hang while installing iptables-persistent on debian 8
+    exec {'iptables-persistent-debconf':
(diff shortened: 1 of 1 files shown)


elyscape:
Could you squash the commits?  Also, does this work on Squeeze and Wheezy or should it only be applied on Jessie?

fsdef:
Sqash done, and it has been tested on wheezy and jessie both work as intended unsure about sqeeze.

fsdef:
Just tested sqeeze, although iptables-persistent doesnt appear to launch debconf there the command will simply add the values to the debconf database.

Still it might be best to restrict it to wheezy/jessie tho to prevent entries without a use.

PR #548, opened 2015-06-19
Don't enable selinux on redhat systems

hunner:


hunner:
We're getting ruby segfaults because of this. Can't do much about that...

ruckc:
No.  WTF? This is horrid.  Why try actually troubleshooting or providing stack trace issue for others to look at, and instead just use the bulldozer to move an anthill.

ruckc:
At least provide a way of letting others troubleshoot your random segfault, instead of just blanket reversing a proven security model for linux.  This is well outside of puppetlabs-firewall's scope of responsibility.

kresss:
This is a terrible plan.  Never mind the fact that the firewall policy has no business in setting the overall SELinux setting for the system.  What if another module foo decided it only works if SELinux is enabled.  

If you need help with SELinux causing a segfault, I am sure there is a community to help with it.

- 1 for this....

jonnytpuppet:
@ruckc @kresss Even though this comment is moot as this change has been reverted, the PR only affected automated beaker test runs, not actual normal use of the module.

PR #550, opened 2015-06-24
Enforce the seluser on selinux systems

hunner:
iptables-save always sets unconfined_u

### DIFF
--- a/manifests/linux/redhat.pp
+++ b/manifests/linux/redhat.pp
@@ -58,10 +58,19 @@ class firewall::linux::redhat (
     require   => File['/etc/sysconfig/iptables'],
   }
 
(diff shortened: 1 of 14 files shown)


manifests/linux/redhat.pp
```diff
@@ -58,10 +58,17 @@
     require   => File['/etc/sysconfig/iptables'],
   }
 
+  # Redhat 7 selinux user context for /etc/sysconfig/iptables is set to unconfined_u
+  case $::selinux {
+    "${::lintminusminus}true",true: { $seluser = 'unconfined_u' }
```
mhaskel: Could we use a [control comment](http://puppet-lint.com/controlcomments/) here instead?

PR #552, opened 2015-06-28
Fix undefined method 'fact' error with Facter 3.x

smoeding:
Using this module with Facter 3.0.1 causes the following error:

```
Error while evaluating a Resource Statement,
Could not autoload puppet/type/firewall:
Could not autoload puppet/provider/firewall/ip6tables:
Could not autoload puppet/provider/firewall/iptables:
undefined method `fact' for Facter:Module
```

It seems that `Factor` no longer has a `fact` method. The fix directly calls the `value` method.

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -841,7 +841,7 @@ Puppet::Type.newtype(:firewall) do
       end
 
       # Old iptables does not support a mask. New iptables will expect one.
-      iptables_version = Facter.fact('iptables_version').value
+      iptables_version = Facter.value('iptables_version')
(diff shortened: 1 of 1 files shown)


johnzimm:
I had the same issue.  I've incorporated this pull request into a local copy of the module and it appears to be working for me now with Facter 3.01 and PuppetServer 4.0.2.

Jmeyering:
:+1:

jonnytpuppet:
It would be appreciated if you could give me your puppet --version and facter --version as I can't actually reproduce this issue.

Thanks

Jmeyering:
Facter Version 3.0.1 (commit 9b1ef723a1494a01e0eccfe93a75fd43316690c2)
Puppet version 4.2.0

jonnytpuppet:
I'm closing this as the .fact method was never removed. You could be hitting another Puppet bug.

PR #558, opened 2015-07-03
Test fixes that may prevent hanging problems during beaker tests.

jonnytpuppet:


spec/acceptance/connlimit_spec.rb
```diff
@@ -9,8 +18,8 @@
           class { '::firewall': }
           firewall { '500 - test':
             proto           => tcp,
-	    dport           => '22',
-	    connlimit_above => '10',
+	          dport           => '2222',
```
tphoney: allignment ?

PR #559, opened 2015-07-14
Compatibility with Puppet 4 and Facter 3

Jmeyering:


mcanevet:
+1

jonnytpuppet:
It would be appreciated if you could give me your puppet --version and facter --version as I can't actually reproduce this issue.

Thanks

mcanevet:
@jonnytpuppet I have the issue with puppet-agent 1.2.1

jonnytpuppet:
I'm using puppet-agent-1.2.1-1.el7.x86_64 on EL7 and don't see this...

Can you provide a manifest that brings up the error please?

Jmeyering:
Facter Version 3.0.1 (commit 9b1ef723a1494a01e0eccfe93a75fd43316690c2)
Puppet version 4.2.0

raphink:
Puppet Agent 1.2.1
Puppetserver 2.1.1

jonnytpuppet:
I'm using the same version and not seeing this.

Can you please confirm if this only happens during the first run? Does it happen after you restart your puppet server?

This could be a known bug in Puppet 4.2.0 that is not related to the firewall module.

jonnytpuppet:
I'm closing this as the .fact method was never removed. You could be hitting another Puppet bug.

peterhuene:
This will manifest only when running under Puppet Server and not an MRI master.  Because Facter 3 is entirely native code, we decided to only implement the minimum set of the Facter API to get it working under a JRuby environment, which is what Puppet Server is.  Thus `Facter.value` is implemented as it is used most frequently and doesn't require a shim "fact" object as an intermediary for retrieving the fact's value.

While `Facter.fact` can technically be used in provider methods as they won't execute on the master, all instances of `Facter.fact` in types and (unfortunately) providers should be replaced with `Facter.value` to be safe.

jonnytpuppet:
Thanks @peterhuene 

I'll submit another PR as the unit tests need updated

raphink:
To be clear, this is what we're seing when using the Firewall module with Puppet 4:

```
Evaluation Error: Error while evaluating a Function Call, Could not autoload puppet/type/firewall: Could not autoload puppet/provider/firewall/iptables: undefined method `fact' for Facter:Module at /etc/puppetlabs/code/environments/staging4/modules/osbase/manifests/firewall.pp:5:5 on node  foo.example.com
```

jonnytpuppet:
https://tickets.puppetlabs.com/browse/MODULES-2246

PR #560, opened 2015-07-15
add set_dscp and set_dscp_class

estonfer:
This adds the ability to make use of --set-dscp and --set-dscp-class flags supported by iptables.  (allowing the setting of DSCP and DSCP Class markings.)


jonnytpuppet:
Thanks for this excellent PR! Happy to merge!

PR #562, opened 2015-07-22
Adding allow ssh rule to tests

jonnytpuppet:
The aim is to keep the beaker connection happy...

### DIFF
--- a/spec/acceptance/rules_spec.rb
+++ b/spec/acceptance/rules_spec.rb
@@ -140,12 +140,20 @@ describe 'complex ruleset 2' do
         ],
       }
 
+      firewall { '001 ssh needed for beaker testing':
+        proto   => 'tcp',
+        dport   => '22',
+        action  => 'accept',
+        before => Firewallchain['INPUT:filter:IPv4'],
+      }
+
       firewall { '010 INPUT allow established and related':
         proto  => 'all',
         state  => ['ESTABLISHED', 'RELATED'],
         action => 'accept',
         before => Firewallchain['INPUT:filter:IPv4'],
       }
+
       firewall { "011 reject local traffic not on loopback interface":
         iniface     => '! lo',
         proto       => 'all',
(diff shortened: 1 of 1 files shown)


spec/acceptance/rules_spec.rb
```diff
@@ -164,6 +164,14 @@ class { '::firewall': }
         action => 'accept',
         before => Firewallchain['INPUT:filter:IPv4'],
       }
+
+      firewall { '22 ssh allow':
```
DavidS: this will not sort correctly as other rules have three digit name prefixes

spec/acceptance/rules_spec.rb
```diff
@@ -257,6 +265,7 @@ class { '::firewall': }
         /-A INPUT -s 10.0.0.0\/(8|255\.0\.0\.0) -p icmp -m comment --comment \"013 icmp echo-request\" -m icmp --icmp-type 8 -j ACCEPT/,
         /-A INPUT -p icmp -m comment --comment \"013 icmp time-exceeded\" -m icmp --icmp-type 11 -j ACCEPT/,
         /-A INPUT -p tcp -m multiport --dports 22 -m comment --comment \"020 ssh\" -m state --state NEW -j ACCEPT/,
+        /-A INPUT -p tcp -m multiport --dports 22 -m comment --comment \"22 ssh allow\" -j ACCEPT/,
```
DavidS: Looking at the line above, that already seems to try to address this, but - judging from the problems we have - fail to do so. Then it can be removed to avoid carrying additional baggage.

Also, it might point to the fact that resetting the FW confuses the netfilter conntrack state that is used by `-m state` being closer to the root cause of all of these issues?

jonnytpuppet:
I agree that we could remove the other rule, but I think that is something that is specifically being tested. I also can't remove the policy => DROP as that is also under test.

spec/acceptance/rules_spec.rb
```diff
@@ -140,12 +140,20 @@ class { '::firewall': }
         ],
       }
 
+      firewall { '001 ssh allow':
```
DavidS: Sorry that I didn't think about it before, but can you add a comment to this rule explaining that it is specifically to allow for the beaker mgmt connection?

```
# allow all traffic from/to the CI-runner host
```

PR #569, opened 2015-08-12
Always use dport

grigarr:
Using 'port' allows you to spoof a source port and offers a backdoor by sending a different destination port. sport 80 -> dport 11211 would be valid for 'port => 80' and give access to a memcache instance (port 11211) running on the server.


xaque208:
+1

PR #572, opened 2015-08-29
MODULES-2487 Improve port deprecation warning

roman-mueller:
When using "port" it currently only throws this warning:

    Warning: port is deprecated and will be removed. Use dport and/or sport instead.

In order to avoid having to grep all your installed module code to figure out which module actually throws this warning: add the module name in the warning.


DavidS:
good idea. thanks!

PR #575, opened 2015-10-30
MODULES-2769 - Add security table for iptables.

werekraken:
This fixes the following error when the built in "security" table is present.

Error: /Stage[main]/Profiles::Firewall::Stateless/Resources[firewallchain]: Failed to generate additional resources using 'generate': Parameter name failed on Firewallchain[INPUT:security:IPv4]: Inbuilt chains must be in the form {chain}:{table}:{protocol} where {table} is one of FILTER, NAT, MANGLE, RAW, RAWPOST, BROUTE or empty (alias for filter), chain can be anything without colons or one of PREROUTING, POSTROUTING, BROUTING, INPUT, FORWARD, OUTPUT for the inbuilt chains, and {protocol} being IPv4, IPv6, ethernet (ethernet bridging) got 'INPUT:security:IPv4' table:'' chain:'' protocol:''

The target system for this fix is CentOS 7, but it should be applicable to any system with a recent kernel/iptables.
https://lwn.net/Articles/267140/
http://www.netfilter.org/projects/iptables/files/changes-iptables-1.4.11.txt

### DIFF
--- a/lib/puppet/provider/firewallchain/iptables_chain.rb
+++ b/lib/puppet/provider/firewallchain/iptables_chain.rb
@@ -38,7 +38,7 @@ Puppet::Type.type(:firewallchain).provide :iptables_chain do
     }
   }
   InternalChains = /^(PREROUTING|POSTROUTING|BROUTING|INPUT|FORWARD|OUTPUT)$/
-  Tables = 'nat|mangle|filter|raw|rawpost|broute'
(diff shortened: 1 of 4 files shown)


jonnytpuppet:
+1

bmjen:
Thanks @werekraken for the contribution! Thanks @jonnytpuppet for taking a look!

PR #576, opened 2015-11-01
Support IPv6 NAT on Linux 3.7+

nward:
This was added to Linux in http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=58a317f1061c894d2344c0b6a18ab4a64b69b815


bmjen:
@jonnytpuppet Do you mind validating this PR?

PR #577, opened 2015-11-04
(MODULES-1341) Recover when deleting absent rules

reidmv:
Some types, specifically the resources type, will call Firewall instances and then use generate to build and add to the catalog firewall resources very early in a Puppet run. Later, those resources might be removed as a side effect of another action, such as shutting down the firewalld service.

Prior to this commit, Puppet would try to delete firewall resources which were already absent, and throw an error. This commit adds an exception catcher which will check to see if the rule being removed is absent, and if so, consider the change a success even if the firewall command failed. It will adjust the change message to reflect the uncertainty over how the rule was removed, though it was verified removed.

### DIFF
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -271,7 +271,22 @@ Puppet::Type.type(:firewall).provide :iptables, :parent => Puppet::Provider::Fir
 
   def delete
     debug 'Deleting rule %s' % resource[:name]
-    iptables delete_args
+    begin
+      iptables delete_args
+    rescue Puppet::ExecutionFailure => e
+      # Check to see if the iptables rule is already gone. This can sometimes
+      # happen as a side effect of other resource changes. If it's not gone,
+      # raise the error as per usual.
+      raise e unless self.resource.property(:ensure).insync?(:absent)
+
+      # If it's already gone, there is no error. Still record a change, but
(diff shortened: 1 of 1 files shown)


tphoney:
@reidmv there is further work required in puppet core, that is due to go into the next release. Until that is done, we cannot merge this PR. As we do not know what the final behaviour will be. Following MODULES-1341 is the best source of information for now.

bmjen:
Hi @reidmv is the explanation from @tphoney sufficient? Can we close this PR?

reidmv:
@bmjen No, I don't think so. I believe @tphoney was referring to PUP-1963 as the work in Puppet core scheduled for the next release, and as I described in MODULES-1341 in [my last comment](https://tickets.puppetlabs.com/browse/MODULES-1341?focusedCommentId=235762&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-235762), my understanding of (and testing with) PUP-1963 indicates that solving PUP-1963 has not solved this issue.

There are three example error messages described in MODULES-1341. PUP-1963 solves the cause of two of them, but not the third.

@bmjen or @tphoney, could one of you arrange for someone to review the comments in MODULES-1341 either to a) clarify what core work you believe will solve this problem, b) to suggest an alternative means of solving this remaining problem, or c) provide feedback on moving this fix forward? Thanks!

PR #578, opened 2015-11-04
(MODULES-2783) Missing ip6tables service name

abednarik:


pshaikh:
We use both iptables & ip6tables, for this we set service_name of firewall class as ['iptables','ip6tables'], the change though is excellent, will it work in situation where one passes service_name as above?

abednarik:
I think the way you were using this module did-t work since they have iptables service hardcoded.

pshaikh:
You're right...iptables was hardcoded, however how to enable both iptables & ip6tables on CentOS or RedHat? By default its always iptables. So what I am doing is passing ['iptables','ip6tables'] for "service_name" attribute of class firewall. This is also done for Gentoo linux. In such a case this may not work.

manifests/linux/redhat.pp
```diff
@@ -46,7 +46,7 @@
       exec { '/usr/bin/systemctl daemon-reload':
         require => Package[$package_name],
         before  => Service[$service_name],
-        unless  => '/usr/bin/systemctl is-active iptables'
+        unless  => '/usr/bin/systemctl is-active iptables',
```
DavidS: shouldn't this also use $service_name?

abednarik:
Thanks @DavidS for providing feedback. Will do that today.

Cheers.

abednarik:
Done @DavidS 

 Any feedback is welcome.

DavidS:
I'll merge it as is. For @pshaikh 's use-case (and I guess this will become more necessary in the coming time) I'd really love to see explicit IPv6 support alongside the current types, so that it can be selected properly at the top-level.

PR #581, opened 2015-11-24
(MODULES-2866) Add: sctp-protocol to "proto"-Parameter

Nold360:
Hi,

this is a simple change, allowing to configure the firewall for the "sctp" Protocol f.e. used by the Pacemaker-Cluster.

Best Regards,
Nold


DavidS:
The test failures are all related to "versioncmp' parameter 'a' expects a String value, got Undef", which is a bug in the testsuite fixed in https://github.com/puppetlabs/puppetlabs-firewall/commit/d27d72f312a32533ef5be15333731816c9bf5178 ; 

If you could rebase on top of current master and push --force here, it should go away and we can merge this then.

Nold360:
Hi,

rebased & test works. :)

DavidS:
Please take a look at https://git-scm.com/book/en/v2/Git-Branching-Rebasing and try again. In the end this PR should only show a single commit that has the current master as single parent. If you need further assistance, I can also do that for you instead.

Nold360:
Sorry, I really don't know how to fix this right now :(

DavidS:
I've rebased this and it is now available at #589 ; I'll merge it as soon as travis is satisfied

PR #582, opened 2015-11-24
Added nodeset files for internal Puppet CI.

jonnytpuppet:


jonnytpuppet:
I think travis is failing due to some other issue

PR #587, opened 2015-11-26
Unit test fixes

jonnytpuppet:


spec/unit/classes/firewall_linux_redhat_spec.rb
```diff
@@ -1,5 +1,27 @@
 require 'spec_helper'
 
+RSpec.shared_examples "common" do
```
DavidS: Since this is in the global scope of all rspec tests, please name this something more specific like `"ensures iptables service"` and use `it_behaves_like "ensures iptables service"` below.

PR #601, opened 2016-01-26
If iptables-save isn't available then assume that no rules are active

epienbroek:
The command 'iptables-save' is marked as being an optional command
so the provider shouldn't produce error messages when it is missing

Resolves the following error message which occurs when iptables-save is
not installed on the system yet and resources { 'firewall': purge => true }
  Error: /Stage[main]/Fw/Resources[firewall]: Failed to generate additional resources using 'generate': Command iptables_save is missing

### DIFF
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -313,8 +313,15 @@ Puppet::Type.type(:firewall).provide :iptables, :parent => Puppet::Provider::Fir
     rules = []
     counter = 1
 
+    # If iptables-save isn't available then assume that no rules are active
+    begin
+      output = iptables_save.split("\n")
+    rescue
+      output = []
+    end
+
     # String#lines would be nice, but we need to support Ruby 1.8.5
-    iptables_save.split("\n").each do |line|
+    output.each do |line|
       unless line =~ /^\#\s+|^\:\S+|^COMMIT|^FATAL/
         if line =~ /^\*/
           table = line.sub(/\*/, "")


tphoney:
@epienbroek 
Thanks for the PR, it looks like this is indeed a problem. Could you please write a test to show this behaviour. Travis.ci is failing too, could you have a look. 

thanks again.

jonnytpuppet:
I'm not sure about this PR, and I'd like a little discussion..

My main concern, is that if the iptables-save command isn't available, then this module will not function, thus any purging will not work properly. So it is probably correct that it errors out, as the module can't do it's job.

I note the way it's marked as optional in the hash in the provider. I'm not sure why this is the case, and I wonder if this is an Puppet API thing that is being used, that in the context of this provider, isn't really optional at all...

jonnytpuppet:
Actually, I just discovered something. The generate() method (used for purging) is always run regardless of the intended order of your resources. Such that if you have a system that doesn't have iptables installed, and include class{'firewall'} to make sure it gets installed, this error will still be thrown. Thus I think this PR (or a variation of it) is warranted.

jonnytpuppet:
I'm still on the fence about this. I'm uncomfortable with allowing this module to be silent when it can't function properly. I'm unsure on how to solve this.

jonnytpuppet:
@epienbroek I'm very sorry, but I think I am going to close this PR. This is part of a bigger issue regarding dependancy/ordering when using dynamically generated resources (such as when using the resources metatype).

In an idea world, we would be able to get 'class{'firewall':}' to run first (which installs the correct packages), before the purging is done using 'resources{'firewall': purge=>true}', however this doesn't seem to be working at present.

I don't believe the solution is to mask the error inside the firewall module, as this will lead to potentially highly undesired behavior where users are unaware of the module working properly in potential other cases.

I appreciate your effort though, and if you have anything more to add, please do let me know.

Thanks

JT

PR #603, opened 2016-02-09
(MODULES 3932) - We need to call Facter.flush to clear Facter cache

jonnytpuppet:
We need to call Facter.flush to clear Facter cache to get up to date value for :iptables_persistent_version.

### DIFF
--- a/lib/puppet/util/firewall.rb
+++ b/lib/puppet/util/firewall.rb
@@ -160,6 +160,9 @@ module Puppet::Util::Firewall
 
     # Older iptables-persistent doesn't provide save action.
     if os_key == 'Debian'
+      # We need to call Facter.flush to clear Facter cache as it's possible the cached value will be nil due to the fact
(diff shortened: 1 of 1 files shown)


tphoney:
@DavidS @hunner is this done in other modules ?

hunner:
It's not, but it sounds valid to me. :+1:

lib/puppet/util/firewall.rb
```diff
@@ -160,6 +160,9 @@ def persist_iptables(proto)
 
     # Older iptables-persistent doesn't provide save action.
     if os_key == 'Debian'
+      # We need to call Facter.flush to clear Facter cache as it's possible the cached value will be nil due to the fact
+      # that the iptables-persistent package was potentially installed after the initial Fact gathering.
+      Facter.flush
```
peterhuene: Rather than flushing all facts, this could potentially be more specific with `Facter.fact(:iptables_persistent_version).flush`, assuming this fact doesn't depend on cached data that is also out of date.

PR #613, opened 2016-03-08
Add the option to add more than 1 firewall rule

cregev:
This will add support for hiera too.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -14,6 +14,7 @@
 class firewall (
   $ensure       = running,
   $pkg_ensure   = present,
+  $rules        = {},
   $service_name = $::firewall::params::service_name,
   $package_name = $::firewall::params::package_name,
 ) inherits ::firewall::params {
@@ -39,4 +40,8 @@ class firewall (
       fail("${title}: Kernel '${::kernel}' is not currently supported")
     }
   }
+  if $rules {
+    create_resources(firewall, $rules)
+  }
+
 }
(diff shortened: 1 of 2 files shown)


metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "puppetlabs-firewall",
-  "version": "1.8.0",
+  "version": "1.8.1",
```
hunner: Please don't bump the metadata as part of a PR; we do that separately.

metadata.json
```diff
@@ -80,6 +80,6 @@
     }
   ],
   "dependencies": [
-  
+      {"name":"puppetlabs/stdlib","version_requirement":">= 4.6.0 <5.0.0"}
```
hunner: Adding a dependency is considered a backwards-incompatible change normally I think, though basically everyone has stdlib installed...

manifests/init.pp
```diff
@@ -14,6 +14,7 @@
 class firewall (
   $ensure       = running,
   $pkg_ensure   = present,
+  $rules        = {},
```
hunner: New parameters would need a documentation update.

jonnytpuppet:
@cregev I'm sorry I don't believe this is the neatest way to solve this problem. What we have done in other modules, is create a separate class for supporting heira. Perhaps you would consider doing the same, rather than adding to the main firewall class? Also, we need a docs update. Thanks!

jonnytpuppet:
Hi There

Thanks for your efforts!

I'm closing this PR due to inactivity. If you would still like to see this feature, feel free to open a new PR with the comments addressed.

Thanks

PR #617, opened 2016-03-14
Add VirtuozzoLinux to the RedHat family

jpnc:
This weekend when we updated the vz packages on our servers, they changed our /etc/redhat-release and some other files to say "VirtuozzoLinux" instead of "CloudLinux", so I'm going through finding places where CloudLinux was hardcoded and fixing all (or at least some of) the things.  Thanks for your consideration!


hunner:
Well that's pretty awkward.

PR #618, opened 2016-03-17
MODULES-1748 Handle negative physdev-is-bridged

kule:
With an iptables line like this:

-A FORWARD -m physdev ! --physdev-is-bridged -j REJECT \
--reject-with icmp-host-prohibited

You'd get:

Parser error: physdev_is_bridged was meant to be a
boolean but received value: REJECT.
- The pre-cludge regex's were being over greedy and taking -j as
  part of the quoted value.
- I allowed @known_booleans to also be "!" - perhaps this could
  be done better tho.
- I also added physdev_is_bridged to the rules that can be prefixed
  with a "!".

### DIFF
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -348,6 +348,8 @@ Puppet::Type.type(:firewall).provide :iptables, :parent => Puppet::Provider::Fir
     values = values.sub(/(!\s+)?--match-set (\S*) (\S*)/, '--match-set "\1\2 \3"')
     # we do a similar thing for negated address masks (source and destination).
     values = values.gsub(/(-\S+) (!)\s?(\S*)/,'\1 "\2 \3"')
+    # catch negated rules with two values
+    values = values.gsub(/(!)\s*(-\S+)\s*-/, '\2 "\1" -')
     # the actual rule will have the ! mark before the option.
     values = values.gsub(/(!)\s*(-\S+)\s*(\S*)/, '\2 "\1 \3"')
     # The match extension for tcp & udp are optional and throws off the @resource_map.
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/firewall/iptables.rb
```diff
@@ -472,7 +474,7 @@ def self.rule_to_hash(line, table, counter)
     # Convert booleans removing the previous cludge we did
     @known_booleans.each do |bool|
       if hash[bool] != nil then
-        if hash[bool] != "true" then
+        if hash[bool] != "true" && hash[bool] != "!" then
```
hunner: I'm not sure what this does... could you express it in a unit test and/or documentation?

kule: Yeah I did wonder if this is worth refactoring more of the code now.  From what I can make out the pre-cludging changes the args e.g.

```! --physdev-is-bridged``` to ```--physdev-is-bridged "!"```  or ```! --state INVALID``` to ```--state "! INVALID"```

Which works for state because it's just basic property however physdev-is-bridged is marked as a boolean so before it would've ended up been hard-coded to a string "true" it now picks up the quoted value in this case "!".

I'll see if I can refactor this a bit more or just add a test - if you have any particular way you'd like this handled let me know.

jonnytpuppet:
@kule thanks for your contribution. Do you know what approach you are taking on this yet? I'd be very interested to hear, thanks!

jonnytpuppet:
Hi There

Thanks for your efforts!

I'm closing this PR due to inactivity. If you would like to see this feature included, feel free to open a new PR.

Thanks!

Flos:
Thanks for the patch, worked for me on CentOS 7.2 with Ovirt.

tphoney:
see #668

PR #626, opened 2016-04-27
Implemented SET jump target to add/del entries to ipsets.

turchanov:


jonnytpuppet:
Thanks for this contribution, we really appreciate your efforts! :)

It would be appreciate if the following points could be addressed before we can merge this:
- Regarding the naming of the new properties, please try and use non-generic terms. For example, "timeout", since it's specifically related to ipset, should probably be something like ipset_timeout. Feel free to disagree with me though if you think I'm wrong :)

- All new properties and parameters need a README update

- All new properties and parameters need some form of testing. I recommend updating the ipset acceptance test to use the new properties.

- Your branch needs rebased against master

Thanks!

turchanov:
> Regarding the naming of the new properties, please try and use non-generic terms. For example, "timeout", since it's specifically related to ipset, should probably be something like ipset_timeout. Feel free to disagree with me though if you think I'm wrong :)

To avoid ambiguity let's discuss all of 4 new properties:
1,2) "add_set", "del_set"  - are not generic, so I leave those names as they are now, ok?
3) "timeout" - this parameter is used in several targets: CT, IDLETIMER, and SET, so that we have to leave that property named a generic way (otherwise we need to invent context-dependant parsing of iptables rules to decide which of "specialized" timeout to assign value to in self.rule_to_hash). Do you agree?
4) "exist" - I will use "set_exist" (to stay in line with "add_set"/"del_set"), ok?

turchanov:
Ping?

tphoney:
Hi @turchanov , thanks for the work you have put into this pull request. However due to inactivity and needing rebased this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request.

PR #629, opened 2016-04-28
() Create a backup of the firewall rules each time puppet makes a change.

heliocentric:
Copy the /etc/sysconfig/iptables file (after stripping out counters)
to a timestamp file in /etc/sysconfig/iptables-backup/, before
running any iptables commands.

Once the firewall provider finishes up, compare the resulting
/etc/sysconfig/iptables file, and if they are different, keep the
backup file. If it's identical, rm the file to save space.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ coverage/
 log/
 .idea/
 *.iml
+*.swp
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -16,18 +16,29 @@ def copy_file(source, destination)
 		f.puts(dest);
 	end
 end
+
 def backup_path(provider)
-  case provider
-  when "iptables"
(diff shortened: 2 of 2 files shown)


jonnytpuppet:
Hi There

Thanks for this contribution, it's much appreciated.

I would suggest you do the following things, before we can merge this:

- Make the functionality optional by adding a parameter to the firewall class. While it's a really cool feature, this isn't something we'd want enabled by default as not all users may want this.
- Add documentation to the parameter that clearly states what this does.
- Tests - We need either (or both!) a spec test or acceptance test to make sure this functionality continues to keep working.

Thanks for your help!

Cheers

JT

tphoney:
Hi @heliocentric , thanks for the work you have put into this pull request. However due to inactivity this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request. Or if you need further assistance reach out to me or @jonnytdevops

PR #630, opened 2016-05-04
(Modules 3329) Add support for iptables length and string extensions

shumbert:
We are using puppetlabs/puppetlabs-firewall in our puppet deployment, however we needed the module to handle the length and string iptables extensions. This pull request contains changes we had to apply.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -410,12 +410,12 @@ This type enables you to manage firewall rules within Puppet.
 
  * `ip6tables`: Ip6tables type provider
     * Required binaries: `ip6tables-save`, `ip6tables`.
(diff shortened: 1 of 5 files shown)


hunner:
Looks good! Could you add a few example lines/hashes to https://github.com/puppetlabs/puppetlabs-firewall/blob/master/spec/fixtures/iptables/conversion_hash.rb so that the tests verify the parameters, and update the readme?

shumbert:
Here is the requested changes.

Originially the string_matching_3 test in ARGS_TO_HASH contained both --from and --to, but it looks like --to is used for both :to and :string_to and the test was failing.

hunner:
👍

PR #637, opened 2016-06-09
(QENG-3973) Adding ubuntu-1604 nodeset

smcelmurry:
Adding these special to this repo due to the ssh timeout used by firewall. Other modules use the qe-ci-module-helper repo for nodesets.


smcelmurry:
@bmjen These are copied from the qe-ci-module-helper repo with the added ssh timeout function

bmjen:
Hmm.. @smcelmurry is this an appropriate fix for the problem? None of the other modules require this nodeset setup.

bmjen:
Nevermind, I see that 2 months ago, we had to override the default nodesets and use the module nodesets in order to set the ssh timeout...

PR #639, opened 2016-06-15
(MODULES-3467) Ignoring rules when purging with firewallchain with use of defined types

glorpen:
This PR simply enables a way to define ignoring rules in multiple places instead of single `firewallchain` with ignore attribute.

Example:

```
firewall::ignore { 'my ignore rules':
        chain => 'INPUT:filter:IPv4',
        regex => ['-j fail2ban-ssh']
}
```

### DIFF
--- a/lib/puppet/type/firewallchain.rb
+++ b/lib/puppet/type/firewallchain.rb
@@ -226,10 +226,27 @@ Puppet::Type.newtype(:firewallchain) do
 
     # Remove rules which match our ignore filter
     rules_resources.delete_if {|res| value(:ignore).find_index{|f| res.provider.properties[:line].match(f)}} if value(:ignore)
+    
+    # Remove rules which match created firewallchain_ignore types
+    catalog.resources.select {
+      |r| r.is_a?(Puppet::Type.type(:firewallchain_ignore)) && r[:chain] == self[:name]
+    }.select {
+      |res| res[:chain] == name
+    }.each { 
+      |ignored_resource| rules_resources.delete_if {|res| ignored_resource[:regex].find_index{|f| res.provider.properties[:line].match(f) }}
+    }
 
     # We mark all remaining rules for deletion, and then let the catalog override us on rules which should be present
     rules_resources.each {|res| res[:ensure] = :absent}
 
     rules_resources
+  end
+  
+  autorequire(:firewallchain_ignore) do
+    catalog.resources.collect do |r|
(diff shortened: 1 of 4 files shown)


cetanu:
This is cool. Why no merge :(

david22swan:
@glorpen Please alter the docs to reflect your changes

david22swan:
Module is currently being put through the rubocop process, will need rebased upon completion in order to merge.

olifre:
I'm also wondering how to actually use this. 

Following the advice from the current `README`, one would define in `site.pp`:
```
resources { 'firewall':
  purge => true,
}
resources { 'firewallchain':
  purge => true,
}
```
I believe the first one has to be dropped when using the `firewall::ignore` mechanism in any case, since otherwise rules with be deleted unconditionally. 

Then, one would define basic rules in `pre` and switch all chains to `DROP` in `post`.
Now, how are unmanaged rules actually dropped?
I can't specify the chain-specific purge:
```
        firewallchain { 'INPUT:filter:IPv4':
                ensure  => present,
                policy  => drop,
                before  => undef,
                purge => true,
        }
```
in `post`, since this would create a dependency cycle, e.g.:
```
Error: Found 1 dependency cycle:
(Firewall[000 accept all icmp] => Class[Profile::Firewall::Post] => Firewallchain[INPUT:filter:IPv4] => Firewall[000 accept all icmp])
```
I can't put it "earlier", though, since this might lead to locking myself out. 

Am I overlooking something? Currently, I'm completely unable to get fail2ban (with f2b-managed rules) to work in combination with `puppet-firewall`, unless I disable the purging functionality completely.

david22swan:
@glorpen Is there any movement on this PR?

glorpen:
@david22swan no, since creating this PR I've simply created own module and moved on as there was more features I've wanted to add/change.
So, for anyone interested, it is now published as https://github.com/glorpen/puppet-g_firewall and soon(ish) on puppet forge.

david22swan:
@glorpen Ah, since that's the case I'll close down this PR. Good luck with your module.

pskopnik:
The generic way to achieve the desired effect with only this module would be the following:

```puppet
firewallchain { 'INPUT:filter:IPv4':
    purge  => true,
    policy => 'drop',
}

# In a different place, e.g. another class:

Firewallchain <| title == 'INPUT:filter:IPv4' |> {
    ignore +> ["-j f2b-ssh"],
}
```

Meta-resource based purging of rules (`resources { 'firewall': purge => true}`) **must be disabled** for this to work.

PR #640, opened 2016-06-23
(MODULES-3514) Also purge fw rules added during catalog apply

jaakkosipari:
See https://tickets.puppetlabs.com/browse/MODULES-3514

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,7 +8,7 @@ PuppetLint.configuration.send('disable_80chars')
 PuppetLint.configuration.send('disable_class_inherits_from_params_class')
 PuppetLint.configuration.send('disable_documentation')
 PuppetLint.configuration.send('disable_single_quote_string_with_variables')
-PuppetLint.configuration.ignore_paths = ["spec/**/*.pp", "pkg/**/*.pp"]
+PuppetLint.configuration.ignore_paths = ["spec/**/*.pp", "pkg/**/*.pp", "vendor/**/*.pp"]
 
 desc 'Generate pooler nodesets'
 task :gen_nodeset do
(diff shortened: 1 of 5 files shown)


jsi-p:
I don't quite understand the UTF-8 error from the build with Travis. The same lint check "works on my machine".

bmjen:
Thanks @jaakkosipari , we're going to have to investigate whether this change will have any unwanted ramifications due to changing the ordering of the code execution from generate to eval_generate

jaakkosipari:
@bmjen I just noticed the change might introduce warnings like this: 'Warning: OUTPUT:filter:IPv4: appears to have a negative number of dependencies'.

I'll try to find out the cause, but it's possible my understanding of Puppet internals and the Firewall module isn't enough to figure out the problem. To reproduce, one probably needs two modules one declaring firewallchain resources and the other (or both) declaring firewall resources for the equivalent chains (eg. INPUT or OUTPUT). It's not necessary to have 'purge => true' for the firewallchain resources to see warnings (while it does give out more 'negative number of dependencies' warnings with purge => true).

Sorry to not have noticed this earlier.

bmjen:
Sounds good! Thanks @jaakkosipari !

jaakkosipari:
I tried to investigate the 'negative number of dependencies'-problem. It seems generate and eval_generate don't just differ on when they're called, but also how the returned data (new resources) is processed and added to the resource graph by Puppet.

Unfortunately I don't have enough time to resolve this issue.

We are just going to resolve our original issue (firewall rules added during puppet run not purged) by using two consecutive puppet runs.

tphoney:
Hi @jaakkosipari , thanks for the work you have put into this pull request. However due to inactivity and needing rebased this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request.

PR #641, opened 2016-06-28
(MODULES-3572) Ip6tables service is not managed in the redhat family.

marcofl:
For the redhat OS family the service for managing ip6tables is called
ip6tables. This service is currently not managed with this module.
This commit fixes this issue by introducing an additional parameter
$service_name_v6 for the ipv6 version of the service.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -2,9 +2,10 @@ class firewall::params {
   $package_ensure = 'present'
   case $::osfamily {
     'RedHat': {
+      $service_name = 'iptables'
+      $service_name_v6 = 'ip6tables'
       case $::operatingsystem {
         'Amazon': {
-          $service_name = 'iptables'
           $package_name = undef
         }
         'Fedora': {
@@ -13,7 +14,6 @@ class firewall::params {
           } else {
             $package_name = undef
           }
-          $service_name = 'iptables'
         }
         default: {
           if versioncmp($::operatingsystemrelease, '7.0') >= 0 {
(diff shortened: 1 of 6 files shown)


marcofl:
looks like some dependency issue in the test environment with "mime-types-data"...

manifests/params.pp
```diff
@@ -5,6 +5,7 @@
       case $::operatingsystem {
```
hunner: You could add the variable assignment above this line, similar to the Debian case below.

hunner:
Thanks for this contribution! Could you also add the `service_name_v6` parameter to the readme, and also an rspec-puppet test similar to the way that gentoo & archlinux do it: https://github.com/puppetlabs/puppetlabs-firewall/blob/1.8.1/spec/unit/classes/firewall_linux_archlinux_spec.rb#L10-L17

For reference, when looking at the way that `service_name` accepts an array of multiple values on gentoo and archlinux would be complicated to apply to redhat due to the sysconfig file resource and service dependency. It's possible, but would require using `join()`, `prefix()` and `create_resources()` :/ (but you can if you want!)

It would also be nice to squash this into one commit, but we can do that too if it's too much trouble. Thanks!

manifests/params.pp
marcofl: done

marcofl:
@hunner  thank you for looking into this. I updated the README, simplified the params for RHEL and copied the correct tests (I hope so) for RHEL. about the array for service_name(_v6): As I cannot think of any reason to have more than one iptables or ip6tables service under RHEL I just put a validate_string to the parameter to ensure no array is supplied. What do you think about this?

tobru:
@hunner Would you mind having a look into this again?

tobru:
Any chance to get this merged soon? It's a rather annyoing issue and it would be great if this issue get's fixed

tphoney:
Thanks for the PR, apologies for taking an age to get at it.

PR #646, opened 2016-08-03
Fixing issue with double quotes being removed when part of the comment

kindred:
If there are double quotes in the comment, the adding of the rule would work fine, but when it came to remove the rule, the double quotes were removed, and the rule was never removed from the system.

This fix will make sure that it only removes the quotes from the beginning and end of the string.


tphoney:
Thanks for the fix. It looks good.

PR #647, opened 2016-08-15
Add --wait to iptables commands

mwhahaha:
This change adds the --wait flag to the insert/update/delete iptables
actions to prevent failures from occuring when iptables is running
outside of puppet. Previously if a script or a user is running iptables
commands while puppet is running, there may be failures in the firewall
provider if iptables cannot get a lock. By default, iptables will just
fail with an error so using --wait will make it wait until a lock can be
established.


mmalchuk:
+1

holser:
+1

tphoney:
Thanks for the great PR ! 👍

PR #649, opened 2016-09-28
iptables_version fact doesn't return the version

pimpolderman:
When running 'facter iptables_version' it doesn't return a value. The bug seems to in the confine not matching the right kernel value.


eputnam:
@pimpolderman thanks for the PR!

PR #650, opened 2016-09-28
ip6tables_version fact doesn't return the version

pimpolderman:
When running the 'facter ip6tables_version' it returns a empty string. The problem is in the confine :kernel => linux line. Adjusted this to capital L.


tphoney:
Nice spot. Thanks for the PR !!!

PR #653, opened 2016-10-17
(maint) Add node yaml file for oracle5

HAIL9000:
Prior to this commit the firewall module did not have a node file
for oracle5 so it was failing in CI. In order to fix this, add the
appropriate config file so tests can be run on oracle5.


hunner:
The module CI jobs actually read from https://github.com/puppetlabs/qe-module-ci-helpers/tree/master/nodes/new and don't use the nodesets in spec/acceptance/nodes. Those are just for our own use.

hunner:
@bmjen points out that firewall DOES at this time use it's own nodesets. My mistake.

PR #656, opened 2016-11-07
enable ip6tables service on Centos/Fedora/Amazon Linux

markwellis:
I noticed on Centos7 that ip6tables isn't started, so the persisted ip6tables rules are not applied at boot.


marcofl:
this duplicates  #641

PR #666, opened 2016-12-08
(MODULES-4200) Add simple sanity check for the rule to hash parser

comel:
We have manual and puppet controlled chains, and someone manually added rule like this:

`-A INPUT -s 1.2.3.4/32 -p tcp -m -tcp --dport 80 --tcp-flags FIN,SYN,RST,ACK SYN -m comment --comment "some comment" -j ACCEPT`

and puppet started throwing errors like:

`Could not evaluate: Invalid address from IPAddr.new: 80`

After some digging I've found that the rule is not properly parsed because of the `--dport 80` between `-m tcp` and `--tcp-flags`. Similar issues can be expected with other combined arguments from the resource_map. It's hard to cover all those cases so I've added simple sanity check for the parser.

https://tickets.puppetlabs.com/browse/MODULES-4200

### DIFF
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -438,10 +438,16 @@ Puppet::Type.type(:firewall).provide :iptables, :parent => Puppet::Provider::Fir
     values.slice!('-A')
     keys << :chain
 
+    valrev = values.scan(/("([^"\\]|\\.)*"|\S+)/).transpose[0].reverse
+
+    if keys.length != valrev.length then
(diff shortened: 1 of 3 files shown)


comel:
Is there anything else I have to do to get this merged or to get some feedback?

jistr:
FWIW the sanity check is the indirect cause of https://tickets.puppetlabs.com/browse/MODULES-6029

Puppet fails to parse some (fairly simple) pre-existing iptables rules. Instead of stopping the Puppet run on this, we should probably just issue a warning and skip the rule.

PR #667, opened 2016-12-13
[#puppethack] MODULES-1222 - added containment

genebean:
This fix seems a little too easy to be all that's needed...


genebean:
Switched to using anchor pattern. Also added stdlib to the metadata and fixtures.

genebean:
back to `contain` and updated metadata to puppet 3.5.0

PR #668, opened 2016-12-13
MODULES-1748 Handle negative physdev-is-bridged

bmjen:
With an iptables line like this:

-A FORWARD -m physdev ! --physdev-is-bridged -j REJECT \
--reject-with icmp-host-prohibited

You'd get:

Parser error: physdev_is_bridged was meant to be a
boolean but received value: REJECT.

* The pre-cludge regex's were being over greedy and taking -j as
part of the quoted value.
* I allowed @known_booleans to also be "!" - perhaps this could
be done better tho.
* I also added physdev_is_bridged to the rules that can be prefixed
with a "!".

Conflicts:
	spec/fixtures/iptables/conversion_hash.rb

### DIFF
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -367,6 +367,8 @@ Puppet::Type.type(:firewall).provide :iptables, :parent => Puppet::Provider::Fir
     end
     # we do a similar thing for negated address masks (source and destination).
     values = values.gsub(/(-\S+) (!)\s?(\S*)/,'\1 "\2 \3"')
+    # catch negated rules with two values
+    values = values.gsub(/(!)\s*(-\S+)\s*-/, '\2 "\1" -')
     # the actual rule will have the ! mark before the option.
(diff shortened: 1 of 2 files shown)


bmjen:
Closes #618

tphoney:
@bmjen can you have a look ?

mhutter:
What's up with this one?

lib/puppet/provider/firewall/iptables.rb
```diff
@@ -367,6 +367,8 @@ def self.rule_to_hash(line, table, counter)
     end
     # we do a similar thing for negated address masks (source and destination).
     values = values.gsub(/(-\S+) (!)\s?(\S*)/,'\1 "\2 \3"')
+    # catch negated rules with two values
+    values = values.gsub(/(!)\s*(-\S+)\s*-/, '\2 "\1" -')
```
mhutter: @bmjen this changes
```
-m addrtype ! --src-type LOCAL
```
to
```
-m addrtype --src "!" -type LOCAL
```

Is this intended?

PR #669, opened 2016-12-14
(FM-5972) gettext and spec.opts

eputnam:


hunner:
https://github.com/puppetlabs/modulesync_configs/pull/114
https://github.com/puppetlabs/modulesync_configs/pull/112

PR #670, opened 2016-12-27
[MODULES-4249] Check status of ebtables package

wiene:
This ensures that firewallchain resources work for the ethernet protocol


manifests/linux.pp
```diff
@@ -27,6 +27,10 @@
     ensure => $pkg_ensure,
   }
 
+  package { 'ebtables':
```
hunner: Could you wrap this in a `$ebtables_manage` class parameter and default to false for users who don't need this package? Thanks!

wiene:
I added the requested parameter.

PR #671, opened 2016-12-28
Add RHEL7 SELinux support for new service_name_v6 param, subsequently fix puppet lint error

wilson208:
This originally started as a fix for a puppet lint issue in CI, where the $seluser variable had been removed from the firewall::linux::redhat class as seluser property for the File["/etc/sysconfig/${service_name}"] resource is now set in a case statement. 

When a separate PR was merged adding the 'service_name_v6' param, it still used the $seluser variable to set the seluser property of the File["/etc/sysconfig/${service_name_v6}"] resource.

When both these PR's were merged they then caused a puppet lint error in CI.

### DIFF
--- a/manifests/linux/redhat.pp
+++ b/manifests/linux/redhat.pp
@@ -72,6 +72,13 @@ class firewall::linux::redhat (
     mode   => '0600',
   }
 
+  file { "/etc/sysconfig/${service_name_v6}":
+    ensure => present,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0600',
+  }
+
   # Before puppet 4, the autobefore on the firewall type does not work - therefore
   # we need to keep this workaround here
   if versioncmp($::puppetversion, '4.0') <= 0 {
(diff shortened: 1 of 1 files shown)


ferventcoder:
I'd prefer to see the additional fields be added as a separate commit than the lint fixes.

wilson208:
@ferventcoder I began with solving the lint issue. However I noticed the lint issue was because the $seluser variable had been removed and the result of solving the lint issue is this PR. I didn't add the additional service_name_v6 param, it was added in PR #641, and the variable $seluser was removed in PR #658. The PR for 641 was created in June and not merged until last week, which is why this was missed when david made his changes in #658 at the end of november.

manifests/linux/redhat.pp
```diff
@@ -85,23 +92,28 @@
         case $::operatingsystemrelease {
           /^7\..*/: {
             case $::operatingsystem {
-              'CentOS': { File["/etc/sysconfig/${service_name}"] { seluser => 'unconfined_u', seltype => 'system_conf_t' } }
-              default : { File["/etc/sysconfig/${service_name}"] { seluser => 'unconfined_u', seltype => 'etc_t' } }
+              'CentOS': {
+                File["/etc/sysconfig/${service_name}"] { seluser => 'unconfined_u', seltype => 'system_conf_t' }
+                File["/etc/sysconfig/${service_name_v6}"] { seluser => 'unconfined_u', seltype => 'system_conf_t' }
```
ferventcoder: This looks like it could be moved to its own commit

ferventcoder:
Now I better understand what is going on after looking closer.

ferventcoder:
Minor nit* - would love to see the git commit message summary be around 50 chars with a message body. I looked over history for this repo and didn't find any long summaries.

PR #672, opened 2017-01-03
Fixes ip6tables errors on RHEL 6/7

wilson208:
Caused through merge of PR's #658 and then subsequently this older PR #641

### DIFF
--- a/manifests/linux/redhat.pp
+++ b/manifests/linux/redhat.pp
@@ -25,8 +25,8 @@ class firewall::linux::redhat (
   # package, which provides the /usr/libexec/iptables/iptables.init used by
   # lib/puppet/util/firewall.rb.
   if ($::operatingsystem != 'Amazon')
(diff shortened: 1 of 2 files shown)


manifests/linux/redhat.pp
```diff
@@ -62,58 +62,62 @@
...
+          $seltype = 'system_conf_t'
+        }
+        default:      {
+          $seluser = 'system_u'
+          $seltype = 'system_conf_t'
+        }
+      }
+    }
+    default:     {}
+    #lint:endignore
```
DavidS: The other changes seem fine, but why does this block have to move?

https://github.com/puppetlabs/puppetlabs-firewall/blob/master/lib/puppet/type/firewall.rb#L1531 will also require updating.

wilson208:
@DavidS I forgot to run spec and lint on this the first time around (oops), I have updated this to make spec tests pass and fix lint errors.

Have addressed your comment on moving that case block, have reverted from using $seluser and $seltype variables to keep order the same. Have also updated the firewall type.

PR #680, opened 2017-01-12
MODULES-4279 Add support for the geoip module

jg-development:
This add the geoip module to the firewall.
I am not very good with ruby, so if somebody can help me out with the spec tests it would be nice.
Except from that, everything works fine


jg-development:
thx

PR #682, opened 2017-01-30
Add gid lookup

crispygoth:
The "uid" parameter for the firewall type correctly deals with mapping between usernames and UIDs as required, however the same isn't true for GIDs, which leads to spurious changes on every puppet run. This PR adds a similar mapping to GIDs.

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -845,6 +845,45 @@ Puppet::Type.newtype(:firewall) do
       only, as iptables does not accept multiple gid in a single
       statement.
     EOS
+    def insync?(is)
+      require 'etc'
+
+      # The following code allow us to take into consideration unix mappings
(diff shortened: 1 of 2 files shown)


crispygoth:
Not sure what docs need updating here, could you point me in the right direction?

As far as I can see the `gid` option is both described in the `README.markdown` already as accepting a "GID or Group owner", mirroring the `uid` option which accepts a "UID or user name". The "forward" mapping of the group name to GID already worked fine. The only thing I've changed with this PR is fixing the "reverse" mapping so that if a group name is specified then it correctly maps the group name in the config to the GID it gets back from the iptables command, which stops it from considering the rule out of sync and trying to "update" the rule every time the agent runs.

david22swan:
Misread the commit, my apologies. Will give the commit a once over the first chance I get.

PR #684, opened 2017-02-09
Allow managing ebtables

hunner:
Closes #670


hunner:
Failure due to https://tickets.puppetlabs.com/browse/PUP-7171 but passes on 4.8.x

PR #686, opened 2017-02-21
Change - Ensure that the iptables configuration file has proper context

blackknight36:
The SELinux context for the iptables configuration file was not being set
properly on Redhat based platforms.  The default context for this file
is 'system_u:object_r:system_conf_t:s0' which is shown by using the restorecon command.


wilson208:
@blackknight36 thanks for this fix. Currently spec tests are failing for this, if you update to get these passing I will run this in our internal adhoc pipeline to ensure it passes on all platforms we test against, if it passes there I am more than happy to merge.

Thanks :)

blackknight36:
I've updated the commit and tests appear to be passing now.

PR #687, opened 2017-02-28
Correction of ICMPv6 protocol name

stefanopini:
Correction of ICMPv6 protocol name from ipv6-icmp to icmpv6


wilson208:
@stefano-93 this fix looks good to me. However you need to update README.md and get the spec tests to pass. After this, I am happy to merge!

Thanks a lot :)

tphoney:
Many thanks for the PR @stefanopini i created #722 with the change to the readme.

PR #685, opened 2017-02-16
(MODULES-4234) Add support for --physdev-is-{in,out}

mhutter:
Fixes MODULES-4234 (see https://tickets.puppetlabs.com/browse/MODULES-4234)

PR #688, opened 2017-03-13
[MODULES-4234] Move physdev_is_in and physdev_is_out tests

wilson208:
Move physdev_is_in and physdev_is_out tests to block with the rest of the 'ip6tables physdev tests' which do not run on EL5 or SLES 10 due to iptables version 1.3.5 is not suppored by the ip6tables provider on those platforms.

This is a result of PR #685 causing CI failures on Redhat 5 and Scientific 5.


pmcmaw:
Looks fine to me. :-)

PR #691, opened 2017-03-22
modulesync e25ca9 - Add locales folder and config.yaml, f4160bf replace version comparisons unnafected by MODULES-4528 to use versioncmp

wilson208:


pmcmaw:
LGTM.

PR #692, opened 2017-03-29
[msync] 786266 Implement puppet-module-gems, a45803 Remove metadata.json from locales config

wilson208:


tphoney:
Closing as it is massively out of date. Other modulesyncs have subsequently happened.

PR #695, opened 2017-04-14
Change - Ensure that firewalld is stopped before iptables starts

blackknight36:
firewalld must be stopped and disabled before the iptables service
will start properly.


blackknight36:
Tests are passing now.

PR #696, opened 2017-05-02
Puppet Lint Fixes

traylenator:


tphoney:
Thanks for the modifications @traylenator !

PR #699, opened 2017-05-07
MODULES-1391 Add u32_matching feature

aaronmoate:
This feature enables the iptables u32 module for making quick and
specific matches against bytes in the packet (i.e. "-m u32 --u32").

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -418,12 +418,12 @@ This type enables you to manage firewall rules within Puppet.
 
  * `ip6tables`: Ip6tables type provider
     * Required binaries: `ip6tables-save`, `ip6tables`.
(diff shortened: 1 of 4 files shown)


hunner:
Thanks for this! Unit tests would be great, and as you mentioned in the ticket some pointers would be helpful. A recently merged pull request that is similar to this one has some unit tests that would be a good example to follow: https://github.com/puppetlabs/puppetlabs-firewall/pull/697/files#diff-4

You can run the tests with `bundle exec rake spec`

eputnam:
@aaronmoate , thanks for your work here! We haven't heard from you in a while so I'm going to close this PR. If you'd still like to get it merged, please feel free to re-open, rebase, and add a few unit tests as mentioned above.

PR #702, opened 2017-05-16
(MODULES-4891) release prep for 1.9.0

eputnam:


eputnam:
@jbondpdx i made some docs changes on here and there's no point in merging them if they're wrong so i thought you could take a look.

PR #721, opened 2017-09-24
(MODULES-5645) Choose correct IP version for hostname resolution

kpengboy:
Currently hostnames specified in a `source` or `destination` field
in a firewall rule are always resolved as IPv4, even when the
provider is `ip6tables`. Instead, intelligently determine whether
the hostname should be resolved as an IPv4 address or IPv6 address
based on the provider.

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -134,8 +134,17 @@ Puppet::Type.newtype(:firewall) do
     EOS
 
     munge do |value|
+      case @resource[:provider]
+      when :iptables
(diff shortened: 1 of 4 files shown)


eputnam:
Thanks @kpengboy !

diranged:
This code actually breaks our Rspec tests - they fail to find the provider. With this in place, is there something I need to do to our spec tests to mock out the provider detection?

eputnam:
@diranged would you mind submitting a ticket for this in JIRA? https://tickets.puppet.com

PR #725, opened 2017-10-04
Revert "Change - Ensure that firewalld is stopped before iptables sta…

blackknight36:
This reverts commit c902b4e609e9d37817dae364728d0506d9178057.

This change creates a dependency cycle which causes puppet agent runs to fail.

### DIFF
--- a/manifests/linux/redhat.pp
+++ b/manifests/linux/redhat.pp
@@ -43,7 +43,7 @@ class firewall::linux::redhat (
     service { 'firewalld':
       ensure => stopped,
       enable => false,
-      before => [Package[$package_name], Service[$service_name]],
+      before => Package[$package_name],
     }
   }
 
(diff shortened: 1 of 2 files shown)


blackknight36:
For example, the following error is shown when using this commit.

Error: Failed to apply catalog: Found 1 dependency cycle:
(Exec[/usr/bin/systemctl daemon-reload] => Service[iptables] => Firewall[000 accept all icmp] => Class[Example::Subsys::Firewall::Pre] => Class[Firewall] => Class[Firewall::Linux] => Package[iptables] => Firewall[000 accept all icmp])
Try the '--graph' option and opening the resulting '.dot' file in OmniGraffle or GraphViz

blackknight36:
Disregard, the issue was in our local manifest, *not* the module itself.

PR #726, opened 2017-10-06
Debian 9 iptables -L returns 2 consecutive spaces

OlivierLM:
Debian 9 iptables -L returns 2 consecutive spaces beetween 'nflog-prefix' and the prefix itself.
The delete_args returns en extra argument, so the rule doesn't get removed.
Matching one space or more in the function regex resolves this problem.

See issue : https://tickets.puppetlabs.com/browse/MODULES-5692


eputnam:
fixed in #727

PR #729, opened 2017-11-03
[MODULES-5924] Fix unmanaged rule regex when updating a iptable.

sathlan:
We have two kind of unmanaged rules:

- empty comment: those get an md5 hash as a name prefixed with 90XX
- comment not matching “\d[graph,space]+$”

This is defined there[1]

But when we calculate the updated line we take into account only the
first kind of unmanaged rule in the offset[2] which leads to error.

Relaxing the regex solves this.

[1] https://github.com/puppetlabs/puppetlabs-firewall/blob/master/lib/puppet/provider/firewall/iptables.rb#L610..L617
[2] https://github.com/puppetlabs/puppetlabs-firewall/blob/master/lib/puppet/provider/firewall/iptables.rb#L802..L803

### DIFF
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -799,7 +799,7 @@ Puppet::Type.type(:firewall).provide :iptables, :parent => Puppet::Provider::Fir
     my_rule = resource[:name].to_s
     rules << my_rule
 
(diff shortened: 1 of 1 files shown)


david22swan:
Will merge once firewall jenkins master goes green

ccamacho:
Any update on this issue?

Thanks!

david22swan:
As above, issues on 2017.3 currently

PR #737, opened 2017-11-22
(MODULES-6092) Set correct seluser for CentOS/RHEL 5.x

mihall-primus:
In 1.10.0, applying the puppetlabs-firewall module on CentOS/RHEL 5.x produces warnings:

Warning: Failed to set SELinux context unconfined_u:object_r:etc_t on /etc/sysconfig/iptables
Notice: /Stage[main]/Firewall::Linux::Redhat/File[/etc/sysconfig/iptables]/seluser: seluser changed 'system_u' to 'unconfined_u'
Warning: Failed to set SELinux context unconfined_u:object_r:etc_t on /etc/sysconfig/ip6tables
Notice: /Stage[main]/Firewall::Linux::Redhat/File[/etc/sysconfig/ip6tables]/seluser: seluser changed 'system_u' to 'unconfined_u'

This commit adds a case to set the correct seluser.

### DIFF
--- a/manifests/linux/redhat.pp
+++ b/manifests/linux/redhat.pp
@@ -106,6 +106,11 @@ class firewall::linux::redhat (
       case $::operatingsystem {
         'CentOS': {
           case $::operatingsystemrelease {
+            /^5\..*/: {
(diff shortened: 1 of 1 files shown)


willmeek:
Many thanks for your contribution, @mihall-primus

PR #738, opened 2017-11-23
(MODULES-6029) Skip unparsable rules with warning

jistr:
The iptables rules parser has very strict/simplistic expectations
about how iptables rules should look like, and can easily fail to
parse rules that weren't produced by the module itself.

We should ignore the unfitting rules when parsing and produce a
warning rather than causing a fatal error and stopping the Puppet run.


pmcmaw:
@jistr thank you for your contribution. 
It is much appreciated 👍

jistr:
Thank you!

pmcmaw:
@jistr If you are waiting on a release of the module please let me know and I will get to work on it asap.

jistr:
@pmcmaw We're trailing master in RDO project trunk repositories, we already have an RPM package generated from the merged commit, so there's no further hurry (we just need to promote trunk repos to stable repos now, which is on the RDO side anyway). Thanks for the quick responses and being helpful when addressing this :)

PR #744, opened 2017-12-07
MODULES-6261: Fix error parsing rules with dashes in the chain name

hantona:
Add a a zero width lookbehind to the regex that processes negations to avoid matching the dash in the chain name.


pmcmaw:
@hantona thank you for taking the time to contribute, it is much appreciated 👍

PR #748, opened 2017-12-20
MODULES-2119 ip6tables delete -p all exception

mikkergimenez:
There is currently a bug in ip6tables which will not match an expression which impacts all L4 protocols, if 'all' as a protocol is not specified (-p all), even though this works with standard version 4 iptables.  This PR works around this bug in ip6tables.  The real fix is in the upstream ip6tables code(bug here:  https://bugzilla.netfilter.org/show_bug.cgi?id=1015)  But this bug has been open for two years.  

Ticket Link: https://tickets.puppetlabs.com/browse/MODULES-2119 (This PR was inspired by code in that ticket)

Without this code, if anyone uses the "all" protocol in an ipv6 rule, the module will be unable to delete the rule. 

I tested this in a vagrant box, by removing and re-adding the conditional, to see if it would fail to delete the rule.


mikkergimenez:
I made a PR that fixes rubocop errors.  Closing this PR.

mikkergimenez:
PR is here now: https://github.com/puppetlabs/puppetlabs-firewall/pull/749

PR #749, opened 2017-12-21
(MODULES-2119) iptables delete -p all exception

mikkergimenez:
This change works around a bug in ip6tables where rules will not be deleted if they attempt to match the 'all' protocol, as it does not properly handle a missing protocol field as an implicit 'all':

netfilter bug located here: https://bugzilla.netfilter.org/show_bug.cgi?id=1015


pmcmaw:
Closing and reopening to resolve Travis/bundler issues, this is caused due to https://tickets.puppetlabs.com/browse/MODULES-6339.

PR #751, opened 2018-01-19
Set static service name

wardhus:
Hi,

I'd like to be able to just notify the service firewall and firewall6 in certain files without checking what OS I'm on. (don't want to duplicate the logic in params.pp)

What do you think of this approach?

### DIFF
--- a/manifests/linux/archlinux.pp
+++ b/manifests/linux/archlinux.pp
@@ -27,10 +27,10 @@ class firewall::linux::archlinux (
   }
 
   service { 'firewall':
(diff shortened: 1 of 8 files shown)


david22swan:
@wardhus This is currently failing on the unit test's. If you could fix the error's that are occurring I would be happy to merge it.

pmcmaw:
Going to close this PR due to inactivity. 
If you have time to pick this up again feel free to reopen.

Also would like to take the time to contribute! 
We really appreciate it 👍

PR #768, opened 2018-08-08
Add support for Amazon Linux 2

erik-frontify:
setting service provider explicit to systemd as pupppets default does
not work for Amazon Linux 2


HelenCampbell:
Thank you for your contribution!

PR #786, opened 2018-10-12
Changed iniface and outiface regex

GeorgeCox:
Changed iniface and outiface regex to allow interface names with an '@' symbol in them. Often used for sub interfaces, i.e vlan10@eth0


david22swan:
@GeorgeCox Please update the reference to reflect your changes. Also it would be preferable if you made your change on a separate branch, rather than on your local copy of the master branch and gave your commit a more descriptive name.

PR #788, opened 2018-10-24
add -g flag handling in ip6tables.rb provider

cestith:
In lib/puppet/provider/firewall/ip6tables.rb there is no goto: entry for the -g flag in @resource_map which leads to errors for rules using that feature. 

lib/puppet/type/firewall.rb already handles this flag. It's a one-line change to recognize it. The right rules then get into the chains in my testing.


cestith:
@tphoney I see you've been active accepting PRs lately. This is a one-line change so I was hoping I could get eyes on it.

tphoney:
Hey @cestith  thanks for the PR, and for having a look at this. Great work !!

PR #791, opened 2018-11-06
Changed regex for iniface and outiface to allow '@' in interface names

GeorgeCox:
Updated regex match for iniface and outiface in lb/puppet/type/firewall.rb
Updated reference in README.markdown to reflect changes in regex.


david22swan:
screen shot 2018-11-06 at 11 59 56 am

PR #793, opened 2018-11-19
Amazon linux 2 changed its major version to 2 with the last update...

erik-frontify:
This fix allows both major versions for Amazon linux 2


HelenCampbell:
Thanks for your contribution!

PR #796, opened 2018-12-11
(MODULES-6340) - Address failure when name begins with 9XXX

eimlav:
When a user specifies a firewall rule starting with a value between 9000-9999, the following error is thrown:

```Error: Could not set 'present' on ensure: undefined method `+' for nil:NilClass at /root/manifest.pp:2```

The following change addresses this by raising a more appropriate error and providing information in the README as to why this error occurs.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -187,6 +187,8 @@ There are two kinds of firewall rules you can use with firewall: default rules a
 
 All rules employ a numbering system in the resource's title that is used for ordering. When titling your rules, make sure you prefix the rule with a number, for example, '000 accept all icmp requests'. _000_ runs first, _999_ runs last.
 
+**Note:** The ordering range 9000-9999 is reserved for unmanaged rules. Do not specify any firewall rules in this range.
+
 ### Default rules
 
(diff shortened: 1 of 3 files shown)


eimlav:
@jonnytdevops Initially I had put the exception in the type however the issue I encountered with this is that because the module assigns unmanaged rules a corresponding firewall type including a title with a value in the range 9000-9999 as intended, this will cause any call to retrieve a firewall resource to fail. This is because Puppet encounters these unmanaged resources and matches their title including the 9000-9999 range as being invalid. Perhaps there would need to be an additional property stating whether the resource refers to an unmanaged rule? What are you thoughts on this?

jonnytdevops:
@eimlav Ah, yes, that is a good point. My suggestion at this point is to somehow tell inside the provider exception if the error is indeed due to a 9000-9999 issue and base the output text on that.

An easier thing you could do is to change the generic error message to provide a indicative advisory (e.g. "Rule sorting error. Make sure that the title of your rule does not start with 9000-9999, as this range is reserved) rather than stating the issue even though we're not sure. It's a "This could be the issue" vs "This is the issue" difference :)

Cheers

JT

PR #797, opened 2018-12-12
(MODULES-8360) Adjust logic to ip6tables-save and restore behavior

glorpen:


glorpen:
I'll add/fix tests after some input from maintainers since this PR could be "me problem" :)

alexharv074:
@glorpen See also https://github.com/puppetlabs/puppetlabs-firewall/pull/804

alexharv074:
@glorpen  Your PR could be a more correct implementation than mine but could you explain the thinking behind it a bit more? My initial impression is that my workaround is only activated in the context of Bug 1015 whereas yours is activated everywhere?

glorpen:
You are right, it is always enabled and no, I don't think its better solution than double deleting as in #804 :)

I've assumed that removing '-p all' would be more correct for newer systems. It does not solve every problem like double deleting seems to.

alexharv074:
Well it looks cleaner but I think it's best to go with my ugly hack! Maybe close this one in favour of #804 ?

alexharv074:
Actually on second thoughts, probably better to leave it open until the maintainers can review them both.

eimlav:
Hey @glorpen . Thanks for creating this PR. After some consideration, we've decided to go ahead with #804 so I will now be closing this ticket. Cheers!

PR #800, opened 2018-12-19
allowing for string_hex type matches for firewall rules

anthonysomerset:
this should fix https://tickets.puppetlabs.com/browse/MODULES-3454 and also allows for creation of string match rules with a hex pattern too


david22swan:
@anthonysomerset Your PR is currently failing against a unit test run. Looking at it it seems to be caused by an issue that has since been fixed and as such should be resolved by a rebase.
In addition in order for your change to be merged it would require test's to be added that cover the feature that you have added.

david22swan:
@anthonysomerset Is there any movement on this pr?

david22swan:
@anthonysomerset Apologies, but due to the amount of time that pr has been left to lie allow and your lack of response when prompted I feel that I must now close this pr. I'm sorry if this inconveniences you and we would off curse be more than happy to review it again should you ever decided to return to it.
Best Wishes

PR #804, opened 2019-01-12
(MODULES-8360) Fix IPv6 bug relating to Bugzilla 1015

alexharv074:
A fix for Bugzilla #1015 was added in
680738164865a01f554d1e1037c8d8724e39a720. However, it appears that the
bug description at https://bugzilla.netfilter.org/show_bug.cgi?id=1015
is not quite accurate, and the -p all hack is required sometimes, but
not always. I don't know exactly when it is or isn't required.

This patch changes the logic to try both.


alexharv074:
Rubocop got me. Have to figure out how to ignore the HandleExceptions check.

alexharv074:
See also https://github.com/puppetlabs/puppetlabs-firewall/pull/797

eimlav:
@alexharv074 Thanks for the contribution! This looks good to me so I'm going to go ahead and merge.

PR #816, opened 2019-02-14
(MODULES-8648) - Fix for failures on SLES 11

david22swan:
SLES 11 Sp4 seems to default to ‘-m conntrack’ rather than ‘-m state’
In addition, adding back a piece of code for Centos 5

### DIFF
--- a/manifests/linux/redhat.pp
+++ b/manifests/linux/redhat.pp
@@ -140,6 +140,11 @@ class firewall::linux::redhat (
         case $::operatingsystem {
           'CentOS': {
             case $::operatingsystemrelease {
+              /^5\..*/: {
+                $seluser = 'system_u'
+                $seltype = 'etc_t'
+              }
+
               /^6\..*/: {
                 $seluser = 'unconfined_u'
                 $seltype = 'system_conf_t'
--- a/spec/acceptance/rules_spec.rb
+++ b/spec/acceptance/rules_spec.rb
@@ -27,7 +27,7 @@ describe 'rules spec' do
           source      => '10.0.0.0/8',
           destination => '!10.0.0.0/8',
           proto       => 'tcp',
-          state       => 'NEW',
+          ctstate       => 'NEW',
           port        => [80,443,21,20,22,53,123,43,873,25,465],
           action      => 'accept',
         }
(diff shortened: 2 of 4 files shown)


david22swan:
screen shot 2019-02-18 at 9 10 21 am

manifests/linux/redhat.pp
```diff
@@ -140,6 +140,11 @@
         case $::operatingsystem {
           'CentOS': {
             case $::operatingsystemrelease {
+              /^5\..*/: {
```
HelenCampbell: Why is this part of this PR? Has nothing to do with SLES 11.

spec/acceptance/rules_spec.rb
```diff
@@ -241,7 +240,7 @@ class { '::firewall': }
     it 'applies cleanly' do
       # Run it twice and test for idempotency
       apply_manifest(pp2, catch_failures: true)
-      apply_manifest(pp2, catch_changes: do_catch_changes)
+      apply_manifest(pp2, catch_changes: true) unless fact('operatingsystem') == 'SLES' && fact('operatingsystemmajrelease') == '11'
```
HelenCampbell: You seem to skip the idempotent checks, is there a reason for this?

manifests/linux/redhat.pp
david22swan: This was removed by accident in an earlier pr

spec/acceptance/rules_spec.rb
david22swan: on sles 11 sp4 State has been deprecated and given a warning to use ctstate instead. It seems to do this for you automatically so everytime the manifest runs it applies as state, but then the firewall changes this to ctstate

manifests/linux/redhat.pp
HelenCampbell: Still has nothing to do with this PR. Either revert the other PR that introduced it or separate it into its own commit. This doesn't make for good git history and would be hard to track.

spec/acceptance/rules_spec.rb
HelenCampbell: I feel like this needs to be a code change instead? If we are passing through the wrong flag we are always going to get those change 'failures', whereas I think the commands being ran need updated to use the -ctstate flag instead of -state?

manifests/linux/redhat.pp
david22swan: Will separate it out into a 2nd commit

spec/acceptance/rules_spec.rb
david22swan: Have been looking at using an if to swap between the two for sless 11.4 but replacing all mentions of state with ctstate might be good as well. They are simply different packages you can choose between. Ctstate just has more features and complexity.

david22swan:
Adhoc running

david22swan:
screen shot 2019-02-18 at 2 40 07 pm

HelenCampbell:
👍

PR #819, opened 2019-02-21
(maint) continued acceptance test cleaning

tphoney:


tphoney:
image
Passing through ad-hoc

PR #826, opened 2019-03-20
(MODULES-8547) Fix existing rules with --dport not parsing

eimlav:


eimlav:
Adhoc ran successfully:

Screen Shot 2019-03-21 at 10 54 42

PR #827, opened 2019-03-21
(MODULES-7333) - Change hashing method from MD5 to SHA256

david22swan:


tphoney:
does this need any modification to the documentation ?

david22swan:
There's no mention of MD5 within the README or comments so I don't think so.

PR #828, opened 2019-03-25
(MODULES-8615) Fix rules with ipvs not parsing

eimlav:

PR #833, opened 2019-04-03
(MODULES-8842) Fix ipvs not idempotent

eimlav:
From my previous PR #828 , it appears that when a rule containing ipvs is applied it is not being saved properly. I have made some additional changes which fix this and also added the missing test.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -524,6 +524,8 @@ If Puppet is managing the iptables or iptables-persistent packages, and the prov
 
 * `bpf`: The ability to use Berkeley Paket Filter rules.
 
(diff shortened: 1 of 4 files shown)


mateusz-gozdek-sociomantic:
Perhaps bugfix release should be done with this then, as support was not properly added in 1.15.2, so it might be confusing for people upgrading.

eimlav:
@mateusz-gozdek-sociomantic I'll get a patch release out as soon as possible.

eimlav:
@mateusz-gozdek-sociomantic 1.15.3 has now been released :)

PR #834, opened 2019-04-04
(MODULES-8855) Move ipvs test to exception spec

eimlav:
ipvs not on RedHat 5/6 by default so moving this to exception spec to fix failing tests.


eimlav:
Adhoc ran successfully
<details>
<summary>Results</summary>
Screen Shot 2019-04-04 at 11 37 11
</details>

PR #844, opened 2019-05-10
Typo fixes

GeorgeCox:
I've just fixed a couple of typos I noticed while using the module, the word 'yphen' where it should be 'hyphen' and the case of the argument error when creating a chain.


eimlav:
Thanks for these fixing these typos @GeorgeCox !

PR #847, opened 2019-05-20
(FM-8025) Add RedHat 8 support

eimlav:
RedHat 8 iptables uses an nftables backend. The `time` match extension
does not appear to be supported. As such, I have moved the test into
exceptions spec.

https://wiki.nftables.org/wiki-nftables/index.php/Supported_features_compared_to_xtables#time

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -19,7 +19,8 @@
       "operatingsystemrelease": [
         "5",
         "6",
-        "7"
+        "7",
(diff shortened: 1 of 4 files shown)


eimlav:
Experimental adhoc including RedHat 8 ran successfully:

Screen Shot 2019-05-20 at 15 57 52

PR #856, opened 2019-09-17
MODULES-9801 - Add functionality to inverse src_type

lionce:


lionce:
not fixing the issue. can be closed

PR #861, opened 2019-10-03
Fix parsing iptables rules with hyphen in comments

Hexta:
Fix parsing rules which have comments including a hyphen mark and no space character.
Example: 101-a-b-c.


Hexta:
Can anybody please take a look at my PR?

weastur:
+
Faced with the same problem.

PR #862, opened 2019-10-08
FM-8400 - add debian10 support

lionce:


lionce:
Pipelines are green
successfully ran on [debian10](https://pipelines.puppet.com/team-modules/builds/594417)

LiohMoeller:
Debian Buster uses nftables by default. In order to be able to use iptables '/usr/sbin/iptables-legacy-save' has to be used. The test pipeline does not reflect the actual production code and will always 'go green' like this.

One way to make it work could be:

```
when :Debian_manual
  case proto.to_sym
  when :IPv4
    if os_key == 'Debian' && Facter.value(:operatingsystem) == 'Debian' && Facter.value(:operatingsystemrelease).to_i >= 10
      ['/bin/sh', '-c', '/usr/sbin/iptables-legacy-save > /etc/iptables/rules']
    else
      ['/bin/sh', '-c', '/sbin/iptables-save > /etc/iptables/rules']
    end
end
```

PR #872, opened 2019-11-08
Support conntrack module

haught:
I would like to add more support for conntrack module options (http://ipset.netfilter.org/iptables-extensions.man.html). These changes do include modifying the existing ctstate option to use the module argument mapping so that the additional options can be added.

### DIFF
--- a/lib/puppet/provider/firewall/ip6tables.rb
+++ b/lib/puppet/provider/firewall/ip6tables.rb
@@ -3,6 +3,7 @@ Puppet::Type.type(:firewall).provide :ip6tables, parent: :iptables, source: :ip6
 
   has_feature :iptables
   has_feature :connection_limiting
+  has_feature :conntrack
   has_feature :hop_limiting
   has_feature :rate_limiting
   has_feature :recent_limiting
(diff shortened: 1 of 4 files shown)


haught:
I have found one oddity I have not figured out yet. If any of the port options are given as the first option: 

```puppet
firewall { '201 test rule':
    chain => 'DOCKER-USER',
    ctorigdstport => '80',
    action => 'drop',
}
```
I get the issue:

> Warning: Puppet::Type::Firewall::ProviderIptables: Skipping unparsable iptables rule: keys (5) and values (7) count mismatch on line: -A DOCKER-USER -p tcp -m conntrack --ctorigdstport 80 -m comment --comment "201 test rule" -j DROP

But the rule created is correct and how I would expect it:

> -A DOCKER-USER -p tcp -m conntrack --ctorigdstport 80 -m comment --comment "201 test rule" -j DROP

If I give any of the options that occur before the port ones in the rule (ctstate, ctproto, ctorigsrc, ctorigdst, ctreplsrc, ctrepldst), everything works. For example:

```puppet
firewall { '201 test rule':
    chain => 'DOCKER-USER',
    ctproto => '6',
    ctorigdstport => '80',
    action => 'drop',
}
```

Is it maybe counting the _ctorigdstport_ option twice, once for _ctorigdstport_ and once for _ctorigdst_ ? But only when first?  It only occurs with ctorigsrcport, ctorigdstport, ctreplsrcport, ctrepldstport as the first option.

haught:
I can get ctorigdst/ctorigdstport et al. working by adding whitespace to the end of the include? comparison in the module argument mapping. There _should_ be a space after every iptables argument it is matching, and it makes it unique. It is not very elegant... is there a better way? It would apply to every module.

```diff
diff --git a/lib/puppet/provider/firewall/iptables.rb b/lib/puppet/provider/firewall/iptables.rb
index c1cba9d..98716f8 100644
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -246,7 +246,7 @@ Puppet::Type.type(:firewall).provide :iptables, parent: Puppet::Provider::Firewa
             resource_map_new[argument][0] = "-m #{ipt_module} #{resource_map_original[argument].first}"
             break
           end
-        elsif compare.include?(resource_map_original[argument])
+        elsif compare.include?(resource_map_original[argument]+' ')
           resource_map_new[argument] = "-m #{ipt_module} #{resource_map_original[argument]}"
           break
         end
```

PR #874, opened 2019-11-12
Change - Avoid puppet failures on windows nodes

blackknight36:
firewall_error
This module will cause puppet agent runs to fail when it is included
on a windows node.  Updated manifest to do nothing when running on
a windows client.


florindragos:
👍

PR #805, opened 2019-01-15
Support for Debian Unstable

pmelse:
adding support for Debian's unstable branch.

note: you may have to purge any rules from iptables-legacy (running "$sudo apt purge iptables-persistent" before re-running puppet.)

I am relatively new to unit testing, so some guidance there would be appreciated if applicable.

PR #876, opened 2019-11-15
Add support for Debian Unstable

martialblog:
Replacement PR for https://github.com/puppetlabs/puppetlabs-firewall/pull/805

 - Added operatingsystemrelease unstable with netfilter service and package
 - Extended rspec tests for Debian

### DIFF
--- a/lib/facter/iptables_persistent_version.rb
+++ b/lib/facter/iptables_persistent_version.rb
@@ -6,7 +6,8 @@ Facter.add(:iptables_persistent_version) do
     os = Facter.value(:operatingsystem)
     os_release = Facter.value(:operatingsystemrelease)
     cmd = if (os == 'Debian' && (Puppet::Util::Package.versioncmp(os_release, '8.0') >= 0)) ||
-             (os == 'Ubuntu' && (Puppet::Util::Package.versioncmp(os_release, '14.10') >= 0))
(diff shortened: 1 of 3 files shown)


florindragos:
Hey @martialblog,
Thank you for your contribution! 👍

LiohMoeller:
The module has been flagged to support also Debian 10 and there are some additions to the spec tests for Buster as well. Actually the module calls iptables-save in the resource definition, which is a symlink to nftables in Debian 10+: https://wiki.debian.org/nftables

Even if the alternatives are configured accordingly afterwards, the module fails on first run with something like:

/Stage[main]/Itsystems_firewall::Config::Request/Resources[firewall] Failed to generate additional resources using 'generate': Execution of '/sbin/iptables-save' returned 1: iptables-save/1.8.2 Failed to initialize nft: Protocol not supported.

There has been some commits addressing the issue: https://github.com/puppetlabs/puppetlabs-firewall/commit/3614e5d3b0ec9a04e34c0e72fc1ae77a4c0c0d91 which are mainly in the acceptance tests and not the actual module which also would need something like this in the resource definition before iptables-save:

```
if os[:family] == 'debian' && os[:release] == '10'
    run_shell('update-alternatives --set iptables /usr/sbin/iptables-legacy')
end
```

PR #886, opened 2020-01-02
(MODULES-10358) - Clarification added to Boolean validation checks

david22swan:


lib/puppet/type/firewall.rb
```diff
@@ -2303,7 +2303,7 @@ def should_to_s(value)
       end
     end
 
-    if value(:log_prefix) || value(:log_level) || value(:log_uid)
+    if value(:log_prefix) == true || value(:log_level) || value(:log_uid) == true
```
DavidS: what happens/should happen when `log_level` is empty string? See https://github.com/puppetlabs/puppetlabs-firewall/blob/5aaedbf9ddb302983fcf0a763f0138362acd0f05/lib/puppet/type/firewall.rb#L754

david22swan: That's handled inside the newproperty level

PR #887, opened 2020-01-06
(maint) tighten down log_level name parsing 

david22swan:


lib/puppet/util/firewall.rb
```diff
@@ -47,8 +47,7 @@ def icmp_name_to_number(value_icmp, protocol)
 
   # Convert log_level names to their respective numbers
   def log_level_name_to_number(value)
-    # TODO: make this 0-7 only
-    if value =~ %r{\d}
+    if value =~ %r{[0-7]}
```
DavidS: This would also allow values lie '000000', '1234567' and 'foo666bar'.

better: `%r{\A[0-7]\z}

david22swan: updated :)

PR #924, opened 2020-06-05
fix parsing of '! --hex-string' from iptables-save

mnejedlo-tds:
iptables-save outputs the value of --hex-string inside quotes, which was
causing quotes inside quotes problems with the value.  This change adds
a special case parser for --hex-string to get the ! inside the quotes
and eliminate the doubled quotation marks

PR #923, opened 2020-06-01
(MODULES-10684) add string_hex to ip6tables provider.

mnejedlo-tds:
pull request #907, applied to ip6tables provider.  MODULES-10684

PR #930, opened 2020-07-16
[IAC-899] - Add acceptance test for string_hex parameter

adrianiurca:
Closes: #923 and closes #924


mnejedlo-tds:
For completeness sake, I should point out that the test included here doesn't test #924. The test for #924 needs to test the negation of --hex-string, e.g. "! --hex-string |fe ed fa ce|", while the test in this PR is for a non-negated --hex-string.

adrianiurca:
Hey @mnejedlo-tds, you're right. Thank you for your observation. I will submit a new PR which includes a new unit test and a new acceptance test(if required). And if something goes wrong we will raise a ticket for corrections.

PR #937, opened 2020-08-05
Fix problem with parsing rules without quotes

scoiatael:
Hello,

we've been using Puppet quite heavily at Opera. Recently I started rolling out fresh version of firewall module on our hosts and noticed a lot of corrective changes being applied, e.g.:
```
Notice: /Stage[main]/Osp_firewall/Firewall[099_a_long_example_rule_with_underscores]/action: current_value , should be 'accept' (noop) (corrective)
```

I think I traced the problem to regex in iptables provider being too greedy and removing suffix with jump target instead of just the name of the rule. I believe the issue is best visible [via regexr](https://regexr.com/59m2h).

Cheers,


adrianiurca:
Hi @scoiatael , thank you for your contribution.

Kind regards,
@adrianiurca

PR #891, opened 2020-01-31
MODULES-10522: Add support for the --condition parameter

mattock:
This allows Puppet to control the entire iptables/ip6tables ruleset while at the
same time allowing some rules to be disabled or enabled outside of Puppet based
on content of files under /proc/net/nf_condition. This feature depends on
xtables-addons.

PR #941, opened 2020-09-02
(IAC-966) - MODULES-10522: Add support for the --condition parameter

adrianiurca:
Closes: #891

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,20 @@ To prevent this issue, do not use MCollective to kick off Puppet runs. Use any o
 * Use a cron job.
 * Click [Run Puppet](https://docs.puppet.com/pe/2016.1/console_classes_groups_running_puppet.html#run-puppet-on-an-individual-node) in the console.
 
+### condition parameter
+
+The `condition` parameter requires `xtables-addons` to be installed locally.
+For ubuntu distributions `xtables-addons-common` package can be installed by running command: `apt-get install xtables-addons-common` or
+running a manifest:
+
+```puppet
+package { 'xtables-addons-common':
+  ensure => 'latest',
+}
+```
+
+For other distributions(RedHat, Debian, Centos etc) is required a manual installation of `xtables-addons` package.
+
 #### Reporting Issues
 
 Please report any bugs in the Puppetlabs JIRA issue tracker:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -474,6 +474,20 @@ To prevent this issue, do not use MCollective to kick off Puppet runs. Use any o
...
+For ubuntu distributions `xtables-addons-common` package can be installed by running command: `apt-get install xtables-addons-common` or
+running a manifest:
+
+```puppet
+package { 'xtables-addons-common':
+  ensure => 'latest',
+}
+```
+
+For other distributions(RedHat, Debian, Centos etc) is required a manual installation of `xtables-addons` package.
```
sanfrancrisko, line 489:
```suggestion
For other distributions (RedHat, Debian, Centos etc) manual installation of the `xtables-addons` package is required.
```

spec/spec_helper_acceptance_local.rb
```diff
@@ -44,15 +44,15 @@ def update_profile_file
 end
 
 RSpec.configure do |c|
+  # This flag is disabling some tests on docker/vagrant containers
+  # To enable tests on abs/vmpooler machines just set to `true` this flag
+  c.filter_run_excluding ubuntu_vmpooler: false
```
sanfrancrisko, line 49: I'm not entirely sure I understand this - the comment states that the flag disables tests on Docker/Vagrant, but is then controlling vmpooler tests?

lib/puppet/provider/firewall/iptables.rb
```diff
@@ -452,6 +455,8 @@ def self.rule_to_hash(line, table, counter)
     values = values.gsub(%r{(!\s+)?--tcp-flags (\S*) (\S*)}, '--tcp-flags "\1\2 \3"')
     # --hex-string output is in quotes, need to move ! inside quotes
     values = values.gsub(%r{(!\s+)?--hex-string "(\S*?)"}, '--hex-string "\1\2"')
+    # --condition output is in quotes, need to move ! inside quotes
+    values = values.gsub(%r{(!\s+)?--condition "(\S*?)"}, '--condition "\1\2"')
```
sanfrancrisko, line 459: You could use `!` to mutate the value of `values`
```suggestion
    values.gsub!(%r{(!\s+)?--condition "(\S*?)"}, '--condition "\1\2"')
```

sanfrancrisko (changes requested):
I would change the title and description of this PR as it is incorporating the work from https://github.com/puppetlabs/puppetlabs-firewall/pull/891 and adding tests to it. You could probably just take the title and description from that PR.

spec/spec_helper_acceptance_local.rb
adrianiurca: exactly

mattock:
Thanks everyone for finishing what I started and failed to complete in time!

PR #944, opened 2020-09-23
Fix extra quotes in firewall string matching

IBBoard:
As reported by Steve Traylon[1] and @patricknelson[2], the munging adds extra quotes to the string.
This breaks the string matching in iptables, as it looks for literal single quotes in the match.

Removing the munging fixes this.

[1] https://tickets.puppetlabs.com/browse/MODULES-3454?focusedCommentId=686988&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-686988
[2] https://github.com/puppetlabs/puppetlabs-firewall/commit/3655c6bd33d662a3813c2f66cd0bc5889c68c2c2#diff-379160b60a9bdf297b92a51d20efd8c3R1426

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -2062,10 +2062,6 @@ Puppet::Type.newtype(:firewall) do
       String matching feature. Matches the packet against the pattern
       given as an argument.
     PUPPETCODE
-
-    munge do |value|
-      _value = "'" + value + "'"
-    end
   end
 
   newproperty(:string_hex) do
--- a/spec/fixtures/iptables/conversion_hash.rb
+++ b/spec/fixtures/iptables/conversion_hash.rb
@@ -1390,7 +1390,7 @@ HASH_TO_ARGS = {
       table: 'filter',
       string: 'GET /index.html',
     },
-    args: ['-t', :filter, '-p', :tcp, '-m', 'string', '--string', "'GET /index.html'", '-m', 'comment', '--comment', '000 string_matching'],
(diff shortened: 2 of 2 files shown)


sanfrancrisko:
Thanks for the fix @IBBoard and references for explanation!

It looks as though [some unit tests are now failing](https://travis-ci.org/github/puppetlabs/puppetlabs-firewall/jobs/729740740#L1579-L1645) as they were testing with the (incorrect!) assertion that the string match values should be wrapped in single quotes. Could you update those tests' data with the new unquoted string values?

I also think [you've highlighted](https://github.com/puppetlabs/puppetlabs-firewall/commit/3655c6bd33d662a3813c2f66cd0bc5889c68c2c2#r42669741) the fact that we don't have an acceptance test that will actually test `iptables` with this flag on real systems. If it has been covered purely by unit tests, and those tests have been built off incorrect assumptions, then it's conceivable this issue was missed. At least with an acceptance test, `iptables` should either blow up or our test expectations would have shown the expected configuration looked incorrect.

IBBoard:
That should have fixed the unit/fixture tests. I've not added any acceptance tests because a) I don't know which of several files they go in and b) from what I understand of them then I can't run them without altering my system, and I don't have anywhere to test them at the moment.

I might be able to write an acceptance test in the dark and hope it works, but it doesn't seem like a good use of Travis cycles!

adrianiurca (approved):

adrianiurca:
Hi @IBBoard , thank you for your contribution.

PR #945, opened 2020-10-02
Update firewall.rb

pellisesol:
On Gentoo using 'vrrp' sets the proto to 112 but when puppet runs it interprets 112 as 'carp' which it then corrects to 'vrrp' which is still 112. So it redoes the rule every time puppet runs. Adding carp here allows carp to be set as a valid option which removes the loop.


adrianiurca (approved):

adrianiurca:
Hi @pellisesol, thank you for your contribution!

PR #948, opened 2020-10-09
(IAC-1190) add `ignore_foreign` when purging firewallchains

DavidS:
This stops `firewallchains` from purging rules that do not look
like rules the module puts into place. This capability can be used
to co-exist with other firewall management solutions on the same
machine.

Improved flow in the README around setting up purging, make it
a user decision to enable, and add a pointer to `ignore_foreign`.


MikaelSmith:
This looks sufficient for our needs.

PR #950, opened 2020-11-16
(IAC-1025) - skip bytecode test on oracle 7

adrianiurca:


spec/spec_helper_acceptance_local.rb
```diff
@@ -43,6 +43,10 @@ def update_profile_file
   run_shell("sed -i '/mesg n || true/c\\test -t 0 && mesg n || true' ~/.profile")
 end
 
+def get_os_name # rubocop:disable Style/AccessorMethodName: Do not prefix reader method names with get_
```
DavidS, line 46: Would recommend using `fetch_` instead of `get_` prefix on any operation that'll take a while.

spec/spec_helper_acceptance_local.rb
```diff
@@ -43,6 +43,10 @@ def update_profile_file
...
+  run_shell('facter os.name').stdout.delete("\n").downcase
```
DavidS, line 47:
```suggestion
  @facter_os_name ||= run_shell('facter os.name').stdout.delete("\n").downcase
```

Add some caching (this won't change during the lifetime of the test runs).

DavidS (commented):
:+1: with two minor comments.

spec/spec_helper_acceptance_local.rb
adrianiurca: done! 👍

PR #952, opened 2020-11-17
pdksync - Update pdk-templates for gitpod and codespaces support

DavidS:
Update pdk-templates for gitpod and codespaces support
pdk version: `1.18.1`


DavidS:
enforce `trusty` dist to avoid docker issues in acceptance tests on travis

PR #971, opened 2021-02-12
Fix build failures caused by safe operator syntax

blackknight36:
The changes from commit 7efed060 cause build failures when running unit tests on CentOS 7.

For example, here is a snippet from our jenkins build logs.

```
02:39:10       Puppet::PreformattedError:
02:39:10         Evaluation Error: Error while evaluating a Resource Statement, Could not autoload puppet/type/firewall: Could not autoload puppet/provider/firewall/ip6tables: /var/lib/jenkins/workspace/mdct-vsftpd/spec/fixtures/modules/firewall/lib/puppet/provider/firewall/ip6tables.rb:71: syntax error, unexpected '.'
02:39:10         ...ptables' if ip6tables_version&.match(%r{1\.3\.\d})
02:39:10         ...                               ^ (file: /var/lib/jenkins/workspace/mdct-vsftpd/spec/fixtures/modules/vsftpd/manifests/init.pp, line: 68, column: 9) on node jenkins.example.com
```

### DIFF
--- a/lib/puppet/provider/firewall/ip6tables.rb
+++ b/lib/puppet/provider/firewall/ip6tables.rb
@@ -68,7 +68,7 @@ Puppet::Type.type(:firewall).provide :ip6tables, parent: :iptables, source: :ip6
 
   def initialize(*args)
     ip6tables_version = Facter.value('ip6tables_version')
-    raise ArgumentError, 'The ip6tables provider is not supported on version 1.3 of iptables' if ip6tables_version&.match(%r{1\.3\.\d})
+    raise ArgumentError, 'The ip6tables provider is not supported on version 1.3 of iptables' if ip6tables_version && ip6tables_version.match(%r{1\.3\.\d})
     super
   end
 
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -651,7 +651,9 @@ Puppet::Type.type(:firewall).provide :iptables, parent: Puppet::Provider::Firewa
         elem.tr(':', '-')
       end
     end
-    hash[:length]&.tr!(':', '-')
+    if hash[:length]
+      hash[:length].tr!(':', '-')
+    end
 
     # Invert any rules that are prefixed with a '!'
     [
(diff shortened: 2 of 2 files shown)


DavidS:
Hi @blackknight36 ,

thanks for your PR. May I ask how you're running those tests?

blackknight36:
> Hi @blackknight36 ,
> 
> thanks for your PR. May I ask how you're running those tests?

We are using rspec-puppet to run unit tests.  Jenkins runs a shell script to look for build failures in each commit.  Here's a copy of the script used by jenkins.

```
#!/bin/bash

export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/opt/puppetlabs/bin:/root/bin

# Script specific for Pull-Request build
if [ -n "$REPO" -a -n "$BRANCH" ]; then
git remote rm proposed || true
git remote add proposed "$REPO"
git fetch proposed
git config --global user.email "michael.watters@dart.biz"
git config --global user.name "Michael Watters"
git checkout origin/production
git merge --no-ff "proposed/$BRANCH" -m "Merge PR"
fi

cd modules/site
rake spec
RESULT=$?
exit $RESULT
```

Essentially this just runs "rake spec" to check for failures.  

Another thing of note is that CentOS 7 provides ruby *2.0* in the default repos which may be why the new syntax is not supported.

```
[root@jenkins ~]# ruby -v
ruby 2.0.0p648 (2015-12-16) [x86_64-linux]
```

DavidS:
> Another thing of note is that CentOS 7 provides ruby 2.0 in the default repos which may be why the new syntax is not supported.

ouch

yeah, we only can support ruby versions that are not yet EOL and that we do ship in our products. I'd recommend using the [PDK rpms](https://puppet.com/docs/pdk/1.x/pdk_install.html) or the [pdk-docker image](https://github.com/puppetlabs/pdk-docker). They also provide the actual bits that we ship in the puppet agent, so you'll get more accurate results.

DavidS:
Closing as the issue is only arising on unsupported ruby versions and install methods.

PR #986, opened 2021-04-15
Update README.md

arjenz:
`state` does not allow inversion, nor in the puppet code in this project nor in iptables rules.


david22swan:
@arjenz Thank you for bringing this to attention.
Hope to see more from you in the future.

PR #1007, opened 2021-08-25
Detect AlmaLinux as RedHat

oranenj:
AlmaLinux seems to be missing from the list of RedHat distros.

Though to be honest, the case statement in linux.pp could probably just check for `$osfamily == "RedHat"` instead of listing all of the options


david22swan:
Closing as support for AlmaLinux has been added, i.e. https://github.com/puppetlabs/puppetlabs-firewall/pull/1020

PR #1013, opened 2021-09-06
Fix rpfilter parameter

onyxmaster:
Hi,
In the current version of the module, the `rpfilter` parameter works as a no-op, no matter which value is provided since the feature is not specified. This commit fixes this issue.

Unfortunately, this is not the only issue with the parameter in question. The [xt_rpfilter](https://ipset.netfilter.org/iptables-extensions.man.html#lbBX) is one of the few extensions that perform meaningful work even when there are no extra parameters specified (in fact the default `strict` mode is invoked by just the `-m rpfilter` stanza), and currently, the Puppet module requires for the mode to be set to one of `loose`/`validmark`/`accept-local`/`invert`. In fact, all of these are optional flags and can be enabled simultaneously or not enabled at all. I'm not sure how to fix this and whether I should report this issue somewhere else. 

So while this PR addresses part of the problem with `rpfilter` parameter just doing nothing, it doesn't fix the main issue with `rpfilter` support. Currently, to use default `strict` matching one can work around the issue by using one of the flags that do not significantly affect the processing most of the time (such as setting `rpfilter => 'validmark'`), but this is only a hacky workaround, not a complete solution. Please advise on where should I report the issue so it could be fixed completely.

### DIFF
--- a/lib/puppet/provider/firewall/ip6tables.rb
+++ b/lib/puppet/provider/firewall/ip6tables.rb
@@ -46,7 +46,6 @@ Puppet::Type.type(:firewall).provide :ip6tables, parent: :iptables, source: :ip6
   has_feature :queue_num
   has_feature :queue_bypass
   has_feature :ct_target
-  has_feature :rpfilter
 
   optional_commands(ip6tables: 'ip6tables',
                     ip6tables_save: 'ip6tables-save')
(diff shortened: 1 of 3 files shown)


onyxmaster:
It looks like the proposed improvement highlighted the issue with testing environment, the CentOS-6 doesn't appear to have support for the `xt_rpfilter` extension :(

onyxmaster:
Well, it looks like I fixed the problem with CentOS 6 by adding proper version checks. @adrianiurca, could you please take a look?

onyxmaster:
This PR is important, but there is no feedback on it :(

LukasAud:
Hi @onyxmaster, sorry for the delay in feedback. Our team has been undergoing some major changes over the last few months and that has translated into a lower level of engagement from our side with the community.

I will remove the 'stale' label and, hopefully, a member from our team might be able to take a look soon at your PR.

LukasAud (approved):

LukasAud:
Hi @onyxmaster, your changes make sense to me and all checks are green, happy to merge this!

Regarding the bigger issue affecting the module, you can report it by creating a new issue here in GH. Please make sure you add as much relevant information as you can (even linking this PR would help) so that the next person that starts working on it has an easier time troubleshooting.

Thanks for your contribution.

PR #1014, opened 2021-09-09
Add carp protocol to :proto property

adrianiurca:
On Gentoo using 'vrrp' sets the proto to 112 but when puppet runs it interprets 112 as 'carp' which it then corrects to 'vrrp' which is still 112. So it redoes the rule every time puppet runs. Adding carp here allows carp to be set as a valid option which removes the loop.

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -554,7 +554,7 @@ Puppet::Type.newtype(:firewall) do
       The specific protocol to match for this rule.
     PUPPETCODE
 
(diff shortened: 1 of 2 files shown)


sanfrancrisko (commented):
Might be worth updating [this test](https://github.com/puppetlabs/puppetlabs-firewall/blob/2aa23fcc86ccd37389b7fafa69a242286753ea3b/spec/unit/puppet/type/firewall_spec.rb#L85-L91) too, as part of the change.

carabasdaniel (approved):
👍  LGTM

sheenaajay (approved):

PR #1016, opened 2021-09-15
Bugfix: use "iptables-compat" on Fedora 34

sircubbi:
Fix module for Fedora 34

see https://tickets.puppetlabs.com/browse/MODULES-11147

### DIFF
--- a/manifests/linux.pp
+++ b/manifests/linux.pp
@@ -44,7 +44,13 @@ class firewall::linux (
     'stopped' => false,
   }
 
-  package { 'iptables':
+  if $::operatingsystem == 'Fedora' and versioncmp($::operatingsystemrelease, '34') >= 0 {
+    $_iptables_pkg = 'iptables-compat'
+  } else{
+    $_iptables_pkg = 'iptables'
+  }
+
+  package { $_iptables_pkg:
     ensure => $pkg_ensure,
   }
 
@@ -66,7 +72,7 @@ class firewall::linux (
         package_name    => $package_name,
         service_name    => $service_name,
         service_name_v6 => $service_name_v6,
-        require         => Package['iptables'],
+        require         => Package[$_iptables_pkg],
       }
     }
     'Debian', 'Ubuntu': {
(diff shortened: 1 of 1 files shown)


adrianiurca (changes requested):
`$::operatingsystem` `$::operatingsystemrelease` are legacy facts and it is possible to not have them in future versions of facter. Could you please change them and use `$facts['os']['name']`, `$facts['os']['release']['full']` in order to have this merged?

kind regards,
@adrianiurca

sircubbi:
Sure, updated all facts, at least in linux.pp.

manifests/linux.pp
```diff
@@ -44,7 +44,13 @@
     'stopped' => false,
   }
 
-  package { 'iptables':
+  if $facts['os']['name'] == 'Fedora' and versioncmp($facts['os']['release']['full'], '34') >= 0 {
+    $_iptables_pkg = 'iptables-compat'
```
bastelfreak, line 48: maybe we should set this in the params.pp? that currently holds the logic to determine data based on the OS

sircubbi: Actually: I am still a bit puzzled why (in the original code) there is a $package_name-variable defined in params.pp, but still the package "iptables" is required hardcoded in linux.pp then.

adamboutcher: I have worked around by doing the following:
```
  ## This is a bodge!! Fedora 34 onwards has no package iptables, its a meta package and breaks the parsing.
  if (($::operatingsystem == 'Fedora' and versioncmp($::operatingsystemrelease, '34') >= 0)) {
    package { 'iptables':
      name   => 'iptables-compat',
      ensure => $pkg_ensure,
    }
  } else {
    package { 'iptables':
      ensure => $pkg_ensure,
    }
  }
```

adamboutcher: However, maybe as @sircubbi says, this should be a param but like this:

```
if this {
$pkg_name = "iptables"
} else {
$pkg_name = "iptables-compat"
}

package { 'iptables':
 name => $pkg_name,
 ensure => $pkg_ensure,
}
```

As this would allow the code to be dropped in but also support different OS package names better in the future.

PR #1023, opened 2021-12-01
fix service port number lookup to use protocol

kjetilho:
The existing code passes `:proto`, which `string_to_port` casts to a
string, gets "proto", compares that to the possibilities "udp" or "tcp",
and when neither, falls back to using "tcp".

This patch passes the actual proto value to the function, in case there is
a UDP specific service in your /etc/services (uncommon, but it happens).
It looks like Puppet will evaluate the properties in declared order,
so I had to move `newproperty(:proto)` up so `@resource[:proto]` was
available in the code for `sport`, `dport` and `port`.

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -382,6 +382,17 @@ Puppet::Type.newtype(:firewall) do
     end
   end
 
+  newproperty(:proto) do
+    desc <<-PUPPETCODE
+      The specific protocol to match for this rule.
+    PUPPETCODE
+
+    newvalues(*[:ip, :tcp, :udp, :icmp, :"ipv6-icmp", :esp, :ah, :vrrp, :carp, :igmp, :ipencap, :ipv4, :ipv6, :ospf, :gre, :cbt, :sctp, :pim, :all].map { |proto|
+      [proto, "! #{proto}".to_sym]
+    }.flatten)
+    defaultto 'tcp'
+  end
+
   newproperty(:sport, array_matching: :all) do
     desc <<-PUPPETCODE
       The source port to match for this filter (if the protocol supports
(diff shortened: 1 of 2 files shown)


kjetilho:
Hrm, looks like the test case "'040 partial invert" needs adjustment?  It now bombs since "http/udp" does not exist in the test harness (it does exist on my Fedora!).  This error was hidden earlier since it looked up "http/tcp".

kjetilho:
This bug is still relevant, and the patch still applies cleanly.

LukasAud:
Hi @kjetilho, thanks for letting us know. We are using the stale-bot as a tool to sort our current PRs and figure out which ones are relevant and which ones might be outdated. We will be putting your PR into the active column and, hopefully, we will be able to review it soon. Thanks for your patience.

LukasAud:
Closing and opening to re-kick automated testing.

LukasAud:
Hi @kjetilho, there seems to be a spec test that is failing currently on your PR. The log is pointing at the following file: 

`rspec ./spec/unit/puppet/provider/iptables_spec.rb:340 # iptables provider when inverting rules fails when not all array items are inverted`

Can you investigate this issue and (if related to your PR) make the necessary changes for the test to pass?

LukasAud:
Hi @kjetilho, there seems to be a spec test that is failing currently on your PR. The log is pointing at the following file: 

`rspec ./spec/unit/puppet/provider/iptables_spec.rb:340 # iptables provider when inverting rules fails when not all array items are inverted`

Can you investigate this issue and (if related to your PR) make the necessary changes for the test to pass?

david22swan:
@kjetilho Any movement on this?

kjetilho:
yep, I can reproduce when testing locally.  will fix soon.  thanks!

jordanbreen28:
@kjetilho Thanks for this - can you rebase with the current main so we can proceed with this.

kjetilho:
not sure if you are notified when I push a rebase, so adding a comment: "sure!" :)

jordanbreen28 (approved):
LGTM - test failures unrelated to PR.

PR #1029, opened 2022-02-17
(SEC-944) Handle duplicate system rules

chelnak:
# Context

In certain situations it is possible for an unmanaged rule to exist on
the target system that has the same comment as the rule specified in
the manifest.

When this condition is true puppet will ignore the the unmanaged rule
and continue to apply the rule in the manifest. This is because the
firewall module uses the comment field in IPT as it's namevar and
therefore expects it to be a unique identifier. In the case of IPT this
is not true given that you can have multiple rules with the same
comment.

# What has changed?

This pr attempts to fix this behaviour by counting the occurrences
of each rule on the system with the expectation that there will only ever
be a count of 1 for each. In the case where the count is greater than 1,
the module will mark the rule as duplicate and emit a warning log
message. When the purge metaparam is set to true, the duplicate rules
will be removed and the puppet agent will reapply the correct one from
the manifest.

### DIFF
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -421,6 +421,10 @@ Puppet::Type.type(:firewall).provide :iptables, parent: Puppet::Provider::Firewa
     rules = []
     counter = 1
 
+    # Create a new hash with a default value of 0. We will store rule counts here
+    # and use it to fish out duplicates.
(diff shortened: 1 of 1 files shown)


david22swan:
Aside from the test failures, the change looks good to me

chelnak:
See #1030 for the latest revision of this change.

PR #1030, opened 2022-02-21
(SEC-944) Handle duplicate system rules

chelnak:
# Context

In certain situations it is possible for an unmanaged rule to exist on the target system that has the same comment as the rule specified in the manifest.

When this condition is true puppet will ignore the the unmanaged rule and continue to apply the rule in the manifest. This is because the firewall module uses the comment field in IPT as it's namevar and therefore expects it to be a unique identifier. In the case of IPT this is not true given that you can have multiple rules with the same comment.

# What has changed?

This commit adds a check that will identify system rules that have their comment field set to the same value as a rule in the manifest. If we enter a situation where any of the duplicate counts are greater than 1 then we will respond with a configurable action. The behaviour of this can be configured via the onduplicaterulebehaviour parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,15 +392,18 @@ firewall {'666 for NFLOG':
 
 ### Duplicate rule behaviour
 
-In certain situations it is possible for an unmanaged rule to exist on the target system that has the same comment as the rule specified in the manifest. This configuration is not supported by the firewall module.
+It is possible for an unmanaged rule to exist on the target system that has the same comment as the rule specified in the manifest. This configuration is not supported by the firewall module.
 
-In the event of a duplicate rule, the module will display a warning message notifying the user that it has found a duplicate but will continue to update the resource.
+In the event of a duplicate rule, the module will by default display a warning message notifying the user that it has found a duplicate but will continue to update the resource.
 
-This behaviour is configurable via the `onduplicaterulebehaviour` parameter. Users can choose from the following actions:
+This behaviour is configurable via the `onduplicaterulebehaviour` parameter. Users can choose from the following behaviours:
 
 * `ignore` - The duplicate rule is ignored and any updates to the resource will continue unaffected.
 * `warn` - The duplicate rule is logged as a warning and any updates to the resource will continue unaffected.
 * `error` - The duplicate rule is logged as an error and any updates to the resource will be skipped.
+
+With either the `ignore` or `warn` (default) behaviour, Puppet may create another duplicate rule.
+To prevent this behavior and report the resource as failing during the Puppet run, specify the `error` behaviour.
 
 ### Additional information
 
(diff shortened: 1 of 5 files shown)


david22swan (dismissed):

lib/puppet/type/firewall.rb
```diff
@@ -234,6 +234,24 @@
...
+      on the target system that has the same comment as the rule
+      specified in the manifest.
+
+      This setting determines what happens when such a duplicate is found.
+
+      It offers three options:
+
+        * ignore - The duplicate rule is ignored and the run will continue.
+        * warn - The duplicate rule is logged as a warning and the run will continue.
+        * error - The duplicate rule is logged as an error and the run will fail.
```
michaeltlombardi, line 249: Suggest clarifying that the _resource_ will be updated (if necessary) vs "the run"

michaeltlombardi (changes requested):
Suggest adding a section to the README for duplicate rules/behavior.

chelnak:
@michaeltlombardi README updated 👍

README.md
```diff
@@ -387,6 +390,18 @@ firewall {'666 for NFLOG':
...
+
+In certain situations it is possible for an unmanaged rule to exist on the target system that has the same comment as the rule specified in the manifest. This configuration is not supported by the firewall module.
+
+In the event of a duplicate rule, the module will display a warning message notifying the user that it has found a duplicate but will continue to update the resource.
+
+This behaviour is configurable via the `onduplicaterulebehaviour` parameter. Users can choose from the following actions:
+
+* `ignore` - The duplicate rule is ignored and any updates to the resource will continue unaffected.
+* `warn` - The duplicate rule is logged as a warning and any updates to the resource will continue unaffected.
+* `error` - The duplicate rule is logged as an error and any updates to the resource will be skipped.
```
michaeltlombardi, line 403:
```suggestion
It is possible for an unmanaged rule to exist on the target system that has the same comment as the rule specified in the manifest. This configuration is not supported by the firewall module.

In the event of a duplicate rule, the module will by default display a warning message notifying the user that it has found a duplicate but will continue to update the resource.

This behaviour is configurable via the `onduplicaterulebehaviour` parameter. Users can choose from the following behaviours:

* `ignore` - The duplicate rule is ignored and any updates to the resource will continue unaffected.
* `warn` - The duplicate rule is logged as a warning and any updates to the resource will continue unaffected.
* `error` - The duplicate rule is logged as an error and any updates to the resource will be skipped.
```

With either the `ignore` or `warn` (default) behaviour, Puppet may create another duplicate rule. To prevent this behavior and report the resource as failing during the Puppet run, specify the `error` behaviour.

michaeltlombardi (dismissed):
Approving the documentation and code implementation, this still needs testing (unit and acceptance).

chelnak:
Tests are currently failing for alma and rocky because of a missing package (``)

With the package installed in the litmusimages, the tests pass as expected. My suggestion would be to fix this in another work item so we can press on with this release.

Here's test evidence for the two OSes with passing acceptance tests.

## AlmaLinux
image

## RockyLinux
image

chelnak:
#1031 raised for unrelated failing tests

david22swan (approved):
LGTM

PR #1048, opened 2022-04-26
fix "Skipping unparsable iptables rule" for rpfilter module

CLNRMN:
Currently, the iptables module rpfilter is not parsed correctly. This leads to it throwing a warning during a puppet run. 

Example of such a warning:
`Warning: Puppet::Type::Firewall::ProviderIptables: Skipping unparsable iptables rule: keys (5) and values (6) count mismatch on line: -A cali-PREROUTING -m comment --comment "cali:mPIOOWmbH3iO0R90" -m mark --mark 0x40000/0x40000 -m rpfilter --validmark --invert -j DROP`

I've fixed the parsing of rpfilter and it's boolean arguments (--loose, --validmark, --accept-local, --invert) according to https://ipset.netfilter.org/iptables-extensions.man.html#lbBX

The change was able to fix the issue in our test environment and the parsing was correct.

For testing I would need some support as I am not quite sure where and how to implement them.

### DIFF
--- a/lib/puppet/provider/firewall/ip6tables.rb
+++ b/lib/puppet/provider/firewall/ip6tables.rb
@@ -151,6 +151,10 @@ Puppet::Type.type(:firewall).provide :ip6tables, parent: :iptables, source: :ip6
     rhitcount: '--hitcount',
     rname: '--name',
     rpfilter: '-m rpfilter',
+    loose: '--loose',
+    validmark: '--validmark',
+    accept_local: '--accept-local',
+    invert: '--invert',
     rseconds: '--seconds',
     rsource: '--rsource',
     rttl: '--rttl',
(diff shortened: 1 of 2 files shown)


chelnak:
Hey 👋 

Thanks for your contribution!

So this PR looks like you are adding functionality that supports rpfilter and it's associated arguments, correct?

If so please could you update your description so that it outlines how this PR intends to fix the problem?

With regards to tests, we are happy to help. Is this something that you are interested in taking for a spin with our support?

CLNRMN:
> So this PR looks like you are adding functionality that supports rpfilter and it's associated arguments, correct?
> 
> If so please could you update your description so that it outlines how this PR intends to fix the problem?

That's correct, I've updated the description. Thanks for your input.

> 
> With regards to tests, we are happy to help. Is this something that you are interested in taking for a spin with our support?

I've asked in the Puppet Community Slack #testing channel for some help regarding testing of this PR.

PR #1061, opened 2022-07-26
allow persistence of firewall rules for Suse

corporate-gadfly:
allow persistence of firewall rules for Suse by using `iptables-save`.


corporate-gadfly:
@david22swan : If you want to give it another whirl, I pushed another commit to possibly fix the Ruby syntax error.

david22swan (approved):
LGTM

Not a complex change and the logic of it looks good so am happy to merge.

Issue #1053, opened 2022-05-20
Puppetlabs-firewall vs Kubernetes takes 10-20+ minutes to parse and add rules


stefanlasiewski:
## Describe the Bug

I'm hoping to get some attention for the bug described at https://tickets.puppetlabs.com/browse/MODULES-10983 . I've been running into it for a couple years. Another person, George Campbell, wrote the bug report MODULES-10983 a year ago, but there hasn't any movement.

In my case, I use Puppet to manage about 150 custom rules (Such as "Allow SSH from the management server to this node" or "Allow the internet to access port 443"). These nodes are part of a Kubernetes cluster.  The CNI used by Kubernetes keeps it's routing rules in IPtables-- there are thousands of rules on every node. 

Here's the problem:

Let's say that I have a new node, or that  I delete all of the rules on my node and run `puppet agent --test` to re-add them. Because of the behavior seen in MODULES-10983, Puppet can take upwards of 30 minutes to write all of the rules.

It would be great to find a way to fix that. 

Is https://tickets.puppetlabs.com/browse/MODULES-10983 a valid place to file bugs or should we file them here instead? If the former, can someone take a look at https://tickets.puppetlabs.com/browse/MODULES-10983?

chelnak:
Hey! Thanks for raising this. It's certainly an interesting issue you are seeing. Having worked on this module I understand what George describes in the MODULES ticket and something I think we should investigate. I can't guarantee a quick fix but I can raise the profile of this ticket within our team.

I'll talk to @Aaron-Rob and get this bumped for community day.

With regards to tracking bugs - here is the right place. We are currently working through some steps to direct our contributors to GitHub issues in the future.

stefanlasiewski:
Great thanks Craig!

Antiarchitect:
Any update on this? We're suffering too.

stefanlasiewski:
Was this rolled out? I don't see the change in https://forge.puppet.com/modules/puppetlabs/firewall/changelog , which as pf today goes up to v4.0.0.

david22swan:
To anyone who is interested in this issue, as part of our investigation into the team has decided that a major rework of the module is needed.
A separate issue has been opened regarding it on which discussion can take place on how we will ultimately proceed: https://github.com/puppetlabs/puppetlabs-firewall/issues/1100

stefanlasiewski:
[...]

PR #1068, opened 2022-08-22
Speedup duplicate rule detection

Antiarchitect:
It seems I've simplified and speed up `duplicate_rule?` function. May be related to https://github.com/puppetlabs/puppetlabs-firewall/issues/1053

### DIFF
--- a/lib/puppet/provider/firewall/iptables.rb
+++ b/lib/puppet/provider/firewall/iptables.rb
@@ -421,11 +421,11 @@ Puppet::Type.type(:firewall).provide :iptables, parent: Puppet::Provider::Firewa
   end
 
   def duplicate_rule?(rule)
-    rules = self.class.instances
-    system_rule_count = Hash.new(0)
-    rules.each { |r| system_rule_count[r.name] += 1 }
-    duplicate_rules = rules.select { |r| system_rule_count[r.name] > 1 }
-    duplicate_rules.select { |r| r.name == rule }.any?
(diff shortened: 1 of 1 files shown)


Antiarchitect:
Will test it in a day or two - we're facing the same issue when number of rules are hundreds

Antiarchitect:
We've tested `duplicate_rule?` and it is called on every `exists?` just to throw some warnings for each rule. So the complexity is O(n^2). We have about 3000 rules on the machine and whole process tooks very long. For the test I just replaced `duplicate_rule?` with

```
def duplicate_rule?(rule)
    false
end
```
 and it gets back to normal. Please do something. It seems we need global hash with O(1) check of a duplicate.

chelnak:
@Antiarchitect We appreciate you going in to such detail with this.

From memory this was a trick one.. due to the requirements around the change O(1) may be hard to achieve.. but i'm certain we can make it faster!

I'll bring it up with the team and get some eyes on it.

stefanlasiewski:
Thanks for looking @chelnak. This is an important fix for us as well.

LukasAud:
Hi @Antiarchitect, sorry for the long delay in feedback. In order to keep better track of this issue and avoid our PR page cluttering, we would like for this topic to be moved into our "Issues" page and, if possible, linked onto our current ongoing discussion about the [Firewall module re-architecture project](https://github.com/puppetlabs/puppetlabs-firewall/issues/1100).

LukasAud:
Closing stale PR. Recommended action has already been stated above.

PR #1080, opened 2022-10-20
(CONT-222) - Adding missing end

pmcmaw:
Prior to this commit a piece of work has been merged into the codebase with a missing end which results in test failures.

This commit adds in the missing end block.


pmcmaw:
the failures were a bigger issue than `end` missing, therefore working on reverting the commit that introduced the failures.

PR #1084, opened 2022-11-03
(CONT-256) Removing outdated code

LukasAud:
Prior to this commit, the firewall modules had multiple instances of outdated code.

This commit aims to clean-up some of that code to keep the module clean and readable.

### DIFF
--- a/manifests/linux/debian.pp
+++ b/manifests/linux/debian.pp
@@ -25,9 +25,11 @@ class firewall::linux::debian (
   $package_name   = $firewall::params::package_name,
   $package_ensure = $firewall::params::package_ensure,
 ) inherits ::firewall::params {
-  ensure_packages([$package_name], {
-      ensure  => $package_ensure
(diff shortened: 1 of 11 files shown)


manifests/linux/debian.pp
```diff
@@ -25,39 +25,16 @@
   $package_name   = $firewall::params::package_name,
   $package_ensure = $firewall::params::package_ensure,
 ) inherits ::firewall::params {
-  if $package_name {
```
david22swan, line 28: You still need the `if $package_name`

david22swan (changes requested):
Oe small comment and a syntax fix.
Otherwise looks good

manifests/linux/debian.pp
LukasAud: re-implemented

david22swan (approved):
LGTM

PR #1085, opened 2022-11-06
package "iptables" has been replaced by "iptables-nft" on EL9

kjetilho:
There are some pointers given by dnf about "iptables", but these confuse Puppet into aborting with the error message:

```console
Error: /Stage[main]/Firewall::Linux/Package[iptables]: Could not evaluate: no implicit conversion of Array into Hash
```

Fedora had a similar patch in commit 486e4b5779f5069e which I think fixed the bug https://tickets.puppetlabs.com/browse/MODULES-11147 but the same issue rared its head here on AlmaLinux 9.0.

The RPM for iptables-legacy states:

> This package contains the legacy tools which are obsoleted by
> nft-variants in iptables-nft package for backwards compatibility reasons.
> If you need to set up firewalls and/or IP masquerading, you should not install
> this package but either nftables or iptables-nft instead.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -33,7 +33,7 @@ class firewall::params {
             $service_name = 'nftables'
             $service_name_v6 = 'ip6tables'
             $package_name = ['iptables-services', 'nftables', 'iptables-nft-services']
-            $iptables_name = 'iptables'
+            $iptables_name = 'iptables-nft'
             $sysconfig_manage = false
             $firewalld_manage = false
           } elsif versioncmp($::operatingsystemrelease, '8.0') >= 0 {


jordanbreen28:
Hi @kjetilho 
Thanks for taking a further look into the test failures!
We are aware that this module has began to fail on el9 and are actively looking into this.
If you feel you may have a suitable fix please do feel free to open a seperate PR. :-)
We appreciate all your work with this!

kjetilho:
EL6 and EL7 do not support `--nflog-size`:

```
       --nflog-range size
              The number of bytes to be copied to userspace (only applicable for nfnetlink_log). nfnetlink_log instances may specify their own range, this  option
              overrides it.
```

kjetilho:
(There must be a keyboard shortcut to close issues inardvertantly?)

I opened PR #1096 for this issue so we can apply the PR's cleanly.

jordanbreen28:
Hi @kjetilho - could you rebase with the current main so the tests rekick & pass, so I can approve :)

jordanbreen28:
Closing & Re-opening to rekick tests

jordanbreen28 (approved):
Approved - thanks again for your hard work @kjetilho. :-)

PR #1088, opened 2022-11-09
(CONT-5) Raising minimum required puppet version

LukasAud:
Prior to this commit, and as part of the codebase hardening project, we sanitised some commands in the module. However, this sanitisation method requires that users have at least puppet version 6.24.0 and onwards to be effective.

This commit aims to resolve that requirement by raising the required puppet version in the metadata.json file.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -79,7 +79,7 @@
   "requirements": [
     {
       "name": "puppet",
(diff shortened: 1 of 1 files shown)


jordanbreen28 (approved):
LGTM

PR #1095, opened 2022-12-02
debian: service: fix `ensure` parameter usage

damonbreeden:
`ensure` is defined but not used anywhere, no clear way to ensure the service is running (and makes it impossible to do so outside the module)


jordanbreen28:
Hi @damonbreeden
Thanks for your work on this one, however it looks as though your change has introduced some spec test failures.
Are you able to address these failures?
Thanks

kjetilho:
A rebase on "main" will probably fix the checks (and trigger a new run of checks)

damonbreeden:
hi, your test failure _expects_ the [service to be unmanaged](https://github.com/puppetlabs/puppetlabs-firewall/actions/runs/3603311805/jobs/6071419351#step:9:1633) but IMO this is an oversight, since `debian` is the only distro that expects this (all others are set to respect the `ensure` param correctly) and since we define `ensure => running` in the [class params](https://github.com/puppetlabs/puppetlabs-firewall/blob/993419129d6451cabf05660dabe44f75d5b79288/manifests/linux/debian.pp#L22) but it's never used anywhere in the class

this makes it impossible to actually manage the running state of the service with puppet, since we define it here as `undef` so we can't define it anywhere else

jordanbreen28:
hi @damonbreeden, thanks for providing some more context surrounding the test failures.

Please feel free to update the debian specific spec tests so that the now pass with this change. (When you have the free time!)

I suspect the failing RHEL-9 issues are due to your branch being outdated, and a fix has since been merged into main for this.

Thanks for your work on this one.

damonbreeden:
hi @jordanbreen28 pls see latest commits, i think they address the spec tests correctly

damonbreeden:
sorry, spoke too quickly
seems like it's failing on an [unknown variable](https://github.com/puppetlabs/puppetlabs-firewall/actions/runs/3857494896/jobs/6575029808#step:9:1233) now, but (should be) unrelated to this PR 
```
       Evaluation Error: Unknown variable: '::osfamily'. (file: /home/runner/work/puppetlabs-firewall/puppetlabs-firewall/spec/fixtures/modules/firewall/manifests/params.pp, line: 7, column: 8) on node fv-az191-465.ggaq42fduhvefhuza3ifwajfkf.bx.internal.cloudapp.net
```

LukasAud (approved):
LGTM

Issue #1097, opened 2022-12-06
Parameter command failed on Exec[/usr/bin/systemctl daemon-reload]


yoshz:
## Describe the Bug
Just upgraded to 4.0.0 and noticed this error:
`Parameter command failed on Exec[/usr/bin/systemctl daemon-reload]`

Seems to be introduced by this commit: https://github.com/puppetlabs/puppetlabs-firewall/commit/dd2cc1eb26d4e50b58738503b8342a0ed94abe1a

I am running Puppet 6 and that doesn't support an array as command.

## Environment
 - Version 4.0.0
 - Platform CentOS 7
 - Puppet 6

LukasAud:
Hi @yoshz, thanks for bringing this to our attention. It seems a mistake was made by reverting the version bump and the module does indeed require a minimum puppet version of 6.24.0 to work properly. We are addressing this issue [here](https://github.com/puppetlabs/puppetlabs-firewall/pull/1098) and a new release should be in the forge soon.

greatflyingsteve:
I am getting a PR prepared which addresses this issue.  The exec in question hasn't been required since at least Puppet 6.11.1 (see PUP-10149), so there is no reason it should have been _changed_ instead of simply _removed_ when the format was converted from command-as-string to command-as-array.  My environment is (for another few weeks) running Puppet 6.19, and this is literally the only thing that was broken for me on pre-6.24 Puppet.

PR #1098, opened 2022-12-06
(GH-1097) Bumping back required puppet version

LukasAud:
Prior to this commit, due to a mistake, there was a reversion to the previously implemented bump to minimum required puppet version, which was part of the CVE update.

This commit aims to address issue 1097 in Github which brings to light the need for this puppet bump to be restored to 6.24


jordanbreen28 (approved):
LGTM

PR #1102, opened 2022-12-13
enable correct service for EL9 using legacy IPv4

kjetilho:
probably not many people still using IPv4 in their systems, after all IPv6 is soon 25 years old, but... we still want to support it.

nftables.service loads nft rules from /etc/sysconfig/nftables.conf, but this module generates classic iptables rules.  The service to load these on boot is simply "iptables.service".

IPv6 rules are loaded correctly by ip6tables.service.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -30,14 +30,14 @@ class firewall::params {
         }
         default: {
           if versioncmp($facts['os']['release']['full'], '9') >= 0 {
-            $service_name = ['nftables','iptables']
+            $service_name = 'iptables'
             $service_name_v6 = 'ip6tables'
             $package_name = ['iptables-services', 'nftables', 'iptables-nft-services']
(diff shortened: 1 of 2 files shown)


LukasAud:
Hi @kjetilho, sorry for the delay in feedback. While reviewing pending PRs for the module, we have merged a different PR that also seems to be addressing this issue, thus causing a conflict. https://github.com/puppetlabs/puppetlabs-firewall/pull/1103

kjetilho:
hmm.  as I mentioned in my comment above, `nftables.service` uses files not managed by this module - so I think it is not correct to enable the service.  who know what rules are left behind in `/etc/sysconfig/nftables.conf`.  I prefer my patch, which only enables `iptables.service`.

kjetilho:
hmm, the "mend" tests failed due to missing api key.  does this mean merge requests need to use project branches rather than external forks?

PR #1104, opened 2023-01-02
Force firewall chain delete

### LINKED ISSUE #1064: Purge rules in firewallchain before removing chain (iptables)
## Use Case

We are using firewall on hosts which may have programs installed which manage their own firewall rules on the host (fail2ban, Docker, ...). `ignore` on `firewallchain` resources is used to allow for these.
Hence, we cannot use meta-resource purging (`resources { 'firewall': purge => true }`) for removing firewall rules and instead we rely on `purge => true` being set on all built-in chains to remove unmanaged and unwanted rules.

A problem arises form these circumstances: Any custom chain disappearing from the Puppet catalog which still contains rules cannot be removed by Puppet.
Both meta-resource purging of chains and declaring the chain as absent fails because there are still rules contained and the iptables provider basically only runs `iptables -X <chain>`

It would be great if there was still a way of removing chains containing rules by first flushing the chain (removing all its rules) and thereafter removing it. These steps should run independently, so that if the first succeeds and the second fails the effect of the first is not rolled back.

I'm not sure what should trigger this behaviour:

 * Meta-resource purging of firewallchain resources: This would constitute a nice counterpart to the meta-resource purging described in the documentation. However, one might argue that purging on firewallchain resources should only delete the chains and not interact with any rules.
 * Declaring a chain as absent: Similar reasoning to above and should probably behave exactly the same as the above.
 * Declaring a chain as absent and setting its `purge => true` parameter: IMO this should certainly cause the rules to be purged before the chain is attempted to be removed. At the moment the chain removal is attempted first and will fail.

Of course having rules in a chain is not the only reason why removal of that chain may fail. The other common reason is that a rule in another chain still links (via a JUMP action) to the to-be-removed chain. However, this is the same scenario also addressed by the meta-resource purging described in the docs and should be handled by several Puppet runs.

There really are two alternatives:

 * Declare the `firewallchain` explicitly but remove the rules via `purge => true`, so only an empty chain remains.
 * Leave it to sysadmins to manually fix the problem (by flushing the chain) in response to Puppet errors.

## Additional Context

\-

cruelsmith:
Flush chain to get rid of unmanaged firewall rules in that chain else the remove of the chain will fail.

---

This PR fixes #1064 where you want to ensure that all unmanaged firewall chains and rules are removed but still want to be able to whitelist unmanaged rules of a chain via `ignore` or `ignore_foreign`.
Code to reproduce:
```puppet
resources { 'firewallchain':
  purge => true,
}

firewallchain { 'INPUT:filter:IPv4':
  purge  => true,
  ignore => [
    '-j fail2ban-', # ignore the fail2ban jump rule
  ],
}

firewallchain { 'fail2ban-sshd:filter:IPv4':
  purge          => true,
  ignore_foreign => true, # ignore not puppet managed rules in this chain
}
```
> **Note**: A additional `resources { 'firewall': purge => true }` is not allowed here in this case because `firewall` will else remove the normally ignored rules by `firewallchain`.
```bash
> puppet agent -t
> iptables -N test
> iptables -I test -j RETURN
> puppet agent -t
```
With this change the second puppet run will not fail and remove every unmanaged chain like this 'test' chain.
> **Note**: If there is still a reference from a rule outside of the chain that should be removed the puppet resource will fail until this reference has been removed. That means there could be still cases where multiple puppet runs are needed.

---

Fixes #1064

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,9 +192,9 @@ resources { 'firewallchain':
 }
 ```
 
-> **Note:** If there are unmanaged rules in unmanaged chains, it will take a second Puppet run for the firewall chain to be purged.
-
 > **Note:** If you need more fine-grained control about which unmananged rules get removed, investigate the `purge` and `ignore_foreign` parameters available in `firewallchain`.
+
+> **Note:** `ignore_foreign` of `firewallchain` does not work as expected with a resources purge of `firewall`.
 
 ### Upgrading
 
--- a/lib/puppet/provider/firewallchain/iptables_chain.rb
+++ b/lib/puppet/provider/firewallchain/iptables_chain.rb
@@ -60,12 +60,14 @@ Puppet::Type.type(:firewallchain).provide :iptables_chain do
   end
 
   def destroy
-    allvalidchains do |t, chain, table|
+    allvalidchains do |t, chain, table, protocol|
       if INTERNAL_CHAINS.match?(chain)
         # can't delete internal chains
         warning "Attempting to destroy internal chain #{@resource[:name]}"
       else
-        debug "Deleting chain #{chain} on table #{table}"
+        debug "Flush chain #{chain} on table #{table} (#{protocol})"
+        t.call ['-t', table, '-F', chain]
+        debug "Deleting chain #{chain} on table #{table} (#{protocol})"
         t.call ['-t', table, '-X', chain]
       end
     end


david22swan (approved):
LGTM

david22swan:
Thx for the fix, noticed this a while back but never got around to doing it myself.

pskopnik:
Thanks @cruelsmith!

PR #1107, opened 2023-01-25
Ignore OpenBSD, similarly to FreeBSD

buzzdeee:
to prevent bailing out in case it gets pulled in by other modules.


david22swan (approved):
lgtm

PR #1123, opened 2023-05-02
Fix SELinux context on newer CentOS

tobias-urdin:
Fix the SELinux contexts for CentOS Stream 8
and CentOS Stream 9.


david22swan (approved):

david22swan:
LGTM thx for the fix

Issue #1100, opened 2022-12-07
Re-Architect Firewall


david22swan:
Hey, just laying this out for everyone so there are no surprises.
As part of a larger discussion, we have decided that the Firewall module is in need of a major restructuring, with more and more issues arising from it and compile times that seem to regularly take upwards of thirty minutes.
In order to try and solve this issue the team is working to put together a plan on how the module should best be restructured to not only resolve the issues currently facing it but to help ensure that similar ones do not occur.
As part of this however we would like to hear the communities feedback and what ideas that they might have on the best way to accomplish this, so that we can reach our goal and provide the best quality module that we can.
As a start there are several questions that must be answered and choices to be made, these including:

* Should we keep to the current provider style module or convert it to run with templates?
Converting the module may take more time to accomplish but could help keep the modules run time down and make the module more easily modified and maintained.

* Should we refactor the current module or create a fresh one and leave the current one as is?
Depending on the size of the refactor it may be easier to simply start fresh, especially if we convert to a template style, as we can leave the old module as is for people to use until we get the new one fully fleshed out.
As a side note however, this could also be accomplished simply by cutting a major release and informing the community beforehand.

* In the event that we create a new module, should we split it into two separate ones, one for iptables and one for ip6tables?
Although there is overlap in how they are managed they are in fact separate and outright keeping them as such may be more convenient for people who simply use on or the other.

* Finally, should the module/modules be expanded to include nftables?
Although it has been around for some time now it is only recently that nftables seems to have gained traction, as such it may be best to add support for it to the firewall module/modules to ease people into converting over, as both it and iptables/ip6tables are both maintained by the same organisation, with the prior being actively developed as a successor and poised to take over.

Finally, to anyone worried about the changes coming to the module, know that this is merely an initial investigation and that the actual changes are still some time away. We are well aware of how important this module is and are committed to getting this right.

This message has also been posted in the puppet community slack and the puppet-users@ mailing list.

canihavethisone:
In a nutshell, i think:

- same module but major version bump
- move to templates,
- Ipv6 in same module
- add nftables so it does everything! (Would you merge in the existing puppet/nftables module logic?)

rtib:
Personal opinion:

[...]

PR #1127, opened 2023-05-12
Add support for parsing and using the SYNPROXY jump target and its options

### LINKED ISSUE #1083: Possible missmatch in rpfilter implementation between iptables and iptables6 providers
Pull-request 048da12b94d37b737a89d54731c7870f227caa39 places `rpfilter` into @known_booleans for [lib/puppet/provider/firewall/ip6tables.rb](https://github.com/puppetlabs/puppetlabs-firewall/blob/3739745e6e98cffa04870280bacf244156b9cd71/lib/puppet/provider/firewall/ip6tables.rb#L234) but doesn't do that for [lib/puppet/provider/firewall/iptables.rb](https://github.com/puppetlabs/puppetlabs-firewall/blob/3739745e6e98cffa04870280bacf244156b9cd71/lib/puppet/provider/firewall/iptables.rb#L231)

As far as I understand the purpose of @known_booleans, I think lib/puppet/provider/firewall/iptables.rb should also include `rpfilter` into @known_booleans (because "-m rpfilter" is a valid match expression for iptables).

Am I wrong?

### LINKED ISSUE #1125: puppetlabs-firewall confused by options used by the SYNPROXY target extension
## Use Case
The puppetlabs-firewall module not only cannot create rules which use the SYNPROXY target extension, it is also confused if it encounters them, and produces a "Skipping unparseable rule" warning.  The failure doesn't prevent other rules from being applied, but they are applied out of order relative to the unparseable rule.  The module also:
- doesn't (can't?) remove the rule
- can't be used to generate the rule
- can't ignore the rule in any way (which also rules out adding it externally)
- forces the rule to the bottom of the chain it's in
- generates (at least) one warning about this same rule for each instance of a Firewall resource that appears in the catalog

The puppetlabs-firewall module should (at a minimum) recognize and (optimally) support the use of options from iptables' SYNPROXY target extension.

The only viable options are: 
- discontinue use of puppetlabs-firewall, and either move to a home-grown solution or attempt to use a different module
- just fix puppetlabs-firewall, which is the route I've taken.

There are too many other good things about the module, and the ability for different teams with entirely unrelated concerns, that coincidentally cut across the same host(s) (e.g. security, application, and operations teams each have competing requirements), to each have their own rule spaces in the codebase that are mediated by the module without needing to worry about who ought to "own" the netfilter subsystem is an enormous organizational win I'm not willing to give up.  

## Additional Context
I'm (currently) using Puppet 6.19 on CentOS 7.  Upgrades to both Puppet 7 and CentOS Stream 9 are in the near future, and the additional overhead of needing to completely reinvent an existing wheel twice in the same year make just about everyone queasy.  As noted in #1097 a bit ago, Puppet 6.24 is not strictly required once the module's hood has been cracked open; despite the recent drop of Puppet 6 from formal support in v5.0.0 of the module, the only actual barrier to using it with Puppet <= 6.11.1 is the systemctl daemon-reload exec that's no longer needed in any case (see PUP-10149).  

SYNPROXY is probably no longer necessary for Linux-based operating systems running 4.4 or newer kernels, as kernels of that release and newer have much better native SYN flood support and don't really need the SYNPROXY target to work around the lack of that kernel code (by adding it in an iptables extension), but any Puppet users making a major OS distribution change are unlikely to put a complete rework of their firewall rulesets in the critical path if the existing rules can be expected to work on their next platform at least during transition.

greatflyingsteve:
Although passing `jump => 'SYNPROXY'` is currently allowed already, the [SYNPROXY target extension](https://ipset.netfilter.org/iptables-extensions.man.html) moves a lot of work normally performed by the kernel's TCP stack out into netfilter, and as such needs to be passed a bunch of options so it knows how to handle the fields and options sent in the initial SYN packet, which are canonically unchangeable once the 3WHS is completed.

This whole arrangement sounds a little crazy, but it's a means of welding the Linux 4.4+ TCP updates that provided better SYN flood protection into older kernels, without incurring the risk of unintended consequences for all the world's existing non-EOL'd, pre-4.4-based, production-facing Linux deployments.  Although that's a dwindling number, there are going to be people migrating to later distro versions that will need to take their existing firewall rulesets with them to decouple OS migration from firewall revision.  As a handy side effect, it can also be used to provide SYN flood protection for other back-end systems receiving forwarded traffic.  The upshot of all this is that this extension is not likely to go away terribly soon.

As an added bonus, this PR also fixes the configuration for the Metrics/BlockLength cop, so that the max block length doesn't need to be constantly hand-incremented each time the Firewall type is given new features.

Additional info in the commit messages, but please ask about anything I forgot to include or messed up!

Resolves #1125.
Resolves #1083.


greatflyingsteve:
Okay so...I'm a little bit worked up right now, and I know (because I too have read #1100) that this is not going to fall on deaf ears, but I'm going to be completely honest: I would have been time, money, and remaining hair ahead to have just ripped the parser completely out and started from scratch implementing it in optparse.  This is _unbelievably bad._  I have never seen Ruby code so unreadable, and as an OSP dev for something like ten years now, this is not a great example of what PuppetLabs can do.

The issue I'm currently solving is that `-m tcpmss --mss <blah>` and `-j SYNPROXY --mss <blah>` need to own and match the `--mss` option, which is what caused the acceptance test failure(s) above; my `:synproxy_mss` was grabbing the `--mss` from `-m tcpmss --mss`, which _helpfully is not even in the unit tests at all_ (until a couple of commits ago).  But while I'm busy solving that, I'm also stumbling right directly across a bunch of cases where this is true, and where the current module structure _is figuratively a time bomb waiting to detonate._  

Assuming that all arguments across the ip[6]tables core and all the extensions are going to be globally unique in a tool that from its inception was based on the idea that an argument's context determines its meaning is the height of naïveté.  Since I only JUST discovered that the target extension options will in some cases sort before some of the matcher extensions in the output of `general_args` (despite that _literally never, in any case,_ happening in the output of `iptables-save`), I'm having to do fun things like read the context entrails to determine if arguments like `--helper` are meant to be `-m helper --helper` or `-j CT --helper` to see whether I need to move that into the pool of _target options_ or the pools of _matcher options._  Split, because I will only sacrifice so much effort without pay, and if I fix that I have to rewrite all the test cases that will fail because of command line argument ordering.  

Then there's also the enjoyable case of `--to` from `-j NETMAP --to` (which internally is just `:to`) and `--to` from `-m string --to` (which internally is at least `:string_to`), both of which simply match with `--to`.  Wanna guess which one you get if you specify that option?  Yeah, me either.  So I'm busy teaching the parser how to be context-aware, and adding a small pool of context data to help it deal with the fact that the correct place to identify an argument for a thing is _in the context block for that thing,_ not literally anywhere in the entire string.

I'm like, this close (`> <`) to just tearing out the entire parsing engine and rewriting it in optparse, which in retrospect, I probably should have done to begin with.  Seriously, why is this 300+ lines of poorly structured regex hell?

greatflyingsteve:
Well, that, um...went places.  Frustration at the failure of acceptance tests I couldn't run and couldn't see without pushing here boiled over into yanking a bunch of the acceptance test cases over into the unit tests, which then quickly boiled over into frustration that I could not run a significant number of those unit tests because I couldn't effectively mock the kernel or ip[6]tables versions that unlocked the provider feature, which then boiled over into frustration that kind of a shocking number of the new unit tests failed immediately (and sometimes violently), which...you get the idea.

...oh, uh, and you can do all of those things now, by the way; I got all of them to actually work, you can mock kernel and iptables versions in iterative test cases, the tests all pass.  That was a fun day-and-a-half foray into the Puppet and RSpec internals, figuring out why my mocked kernel version still wouldn't unlock `has_feature :rpfilter` (and _then_ figuring out why it was still broken for ip6tables once the feature was available!).

There are still a fair few features that turn out to be miss-implemented, most of which I've noted in the commit messages where I partially-fixed them, which I still kinda want to get my grubby li'l paws on. But as [@bastelfreak pointed out last week](https://github.com/puppetlabs/puppetlabs-firewall/issues/1100#issuecomment-1552870577), it's not exactly worth the effort, and my employer at this point is probably pretty desperate for me to move on to other things.

greatflyingsteve:
Also: I'm happy to split this into multiple PRs if necessary, but that comes with the caveat that tests in this PR will appear to fail until after a few of the other fixes that make parsing the `--mss` argument differently for `-m tcpmss --mss` than for `-j SYNPROXY --mss` are in place.

david22swan:
@greatflyingsteve Apologies for the wait on having this looked at.
This looks like a good change to me, however the module is currently undergoing work to convert it to the resource_api so this will likely reworked before it can be merged.
Once again, sorry for the delay.

jst-cyr:
Hello @greatflyingsteve , I'm doing a pass through all the old PRs to try to get a sense of what still needs addressing. I saw this PR mentioned on #1145 by you. Would you consider the needs of this PR met? Can we close this one?

jst-cyr:
Given changes in #1245 and the age of the PR, I'm going to close this one for now, but please feel free to re-open!

PR #1134, opened 2023-05-25
Change ip6tables_version to constant in provider.

pjakubcz:
Facter is being asked for ip6tables_version every time ip6tables provider constructor is called during prefetch. This causes severe performance issues on systems with a lot of ip6tables rules (fe: openstack hypervisors) when even a single ipv6 rule is added. This commit should fix this.


david22swan (approved):
LGTM

PR #1135, opened 2023-05-30
pdksync - (MAINT) - Allow Stdlib 9.x

LukasAud:
(MAINT) - Update Stdlib dependency
pdk version: `2.7.1`

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.0.0 < 9.0.0"
+      "version_requirement": ">= 4.0.0 < 10.0.0"
     }
   ],
   "operatingsystem_support": [
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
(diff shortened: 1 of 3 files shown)


pmcmaw:
Resolves #1137

metadata.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.0.0 < 9.0.0"
+      "version_requirement": ">= 4.0.0 < 10.0.0"
```
bastelfreak, line 13: the namespaced functions aren't available in 8.6:
```suggestion
      "version_requirement": ">= 9.0.0 < 10.0.0"
```

LukasAud: This makes sense. I will probably have to revisit some of the other PRs/modules (APT probably) to make sure any module affected by namespaces is properly bumped. Its a fairly large bump (from 4 to 9), I wonder if it could have any side effects im not seeing here. Like the issue we run into when we tried to set up TargetRubyVersion to 2.7.

bastelfreak: because of that we use a more meaningful PR title at vox pupuli. Usually: `puppetlabs/stdlib: Require 9.x` or `puppetlabs/stdlib: Allow 9.x`

PR #1139, opened 2023-06-07
(CONT-242) Fix duplicate rule detection

david22swan:
This was previously accomplished by retrieving the full list of rules each time a rule was set in order to check it was unique. This was to allow the user to choose the response to a duplicate rule being found.

However this caused a massive slowdown within certain module runs and as such we have changed the the location of the check, so that it instead runs a check for any duplicates when retrieving the current rules prior to any updates being made.

As an effect of this the user is now unable to choose the response to a duplicate rule being found, however wee feel that this is a fair tradeoff for the increased speed and that the response that we have chosen is the correct one.

### DIFF
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -236,24 +236,6 @@ Puppet::Type.newtype(:firewall) do
     newvalues(%r{^\d+[[:graph:][:space:]]+$})
   end
 
-  newparam(:onduplicaterulebehaviour) do
-    desc <<-PUPPETCODE
-      In certain situations it is possible for an unmanaged rule to exist
-      on the target system that has the same comment as the rule
-      specified in the manifest.
-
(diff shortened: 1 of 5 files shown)


lib/puppet/type/firewall.rb
```diff
@@ -236,24 +236,6 @@
     newvalues(%r{^\d+[[:graph:][:space:]]+$})
   end
 
-  newparam(:onduplicaterulebehaviour) do
```
bastelfreak, line 239: is that actually a bugfix? Well it might fix a bug, but it's also a breaking change, isn't it?

david22swan: Probably yeh

david22swan:
Recreated as: https://github.com/puppetlabs/puppetlabs-firewall/pull/1140
The PR somehow un-synced from the branch????

PR #1142, opened 2023-07-02
disable firewalld for RedHat 9

robertc99:
## Summary
disable firewalld for redhat 9

## Additional Context
firewalld and iptables services conflict


jordanbreen28 (approved):
Nice one @robertc99! 
Thanks for fixing this.

PR #1145, opened 2023-07-18
(DO NOT MERGE)(CAT-376) Rework firewall module to use the resource_api

david22swan:
## Summary
This PR has been created to update the Firewall module to utilize the resource_api, with the intention being to increase the ease with which the module can be updated and managed in the future.
As part of this several backwards incompatible changes where made, including:

1. The `provider` attribute with the Firewall type has been renamed `protocol`, due to both the resource_api forbidding the use of this as an attribute name and to bring it in line with the Firewallchain. Though it continues to accept both `iptables` and `ip6tables` alongside `IPv4` and `IPv6`, this may change in the future.
2. The `action` attribute has been removed as it managed the same function as the `jump` attribute, only on a more limited scale. Though this was given a reason, to enforce the use of generic parameters, I found this to be a needlessly complex addition to the code.
3. Strict types have been declared for all attributes.
4. The `port` attribute has been removed as it was deprecated several years ago.
5. `dport`/`sport` ranges should be passed with `:` as a separator as this is the valid form of input for the flag. _Code has been updated to allow the original separator `-`, though it is not preferred._
6. Array attributes, such as `sport`/`dport`, that require negation to be universal among all passed values have been updated so that rather than negating each and every value passed you now simply negate the first value in the array in order to negate them all, although the original form of negating each and every value is still allowed. _Certain array values such as `src_type`/`dst_type` differ however and necessitate that each value be negated, or not negated as it were, separately._

**NOTES: 

- firewallchain purge tests disabled, waiting on `puppet-resource_api` functionality to be merged and released: https://github.com/puppetlabs/puppet-resource_api/pull/316 --> `puppet-resource_api` change has been merged and released (v1.9.0), waiting on update to `puppet-agent` to include this new version so that I can turn on the automated tests https://github.com/puppetlabs/puppet-agent/pull/2383 --> PR has been merged and will be included in `7.26.0` and `8.2.0` --> tests enabled
- Unable to use the Stdlib IP types for most address style input as they include extra features such as ranges and negation.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.


lib/puppet/provider/firewall/firewall.rb
greatflyingsteve, line 1: After the struggle I went through to get #1127 in working order, there are large parts of this file that read like a bandaid being torn off quickly.  It might not be fun to re-work my PR to fit this framework, but I'm betting it'll be _so much easier_ than writing the original draft was.  Reading through this version is so much more satisfying!

One question, though - if the output of `ip[6]tables-save` is just the arguments one should append to `iptables -t <blah>` to regenerate the same rule again, would it have been simpler to use [optparse](https://ruby-doc.org/stdlib-2.7.0/libdoc/optparse/rdoc/index.html) to recover the sense from the rules' arguments?

david22swan: Thanks for the positive response :)

In regards to the use of `optparse`, I had not encountered it before and so didn't consider it as an option.
It may be a little late for me to be rewriting the code, I am scheduled to move onto other work, however I will make sure to ticket it up as an improvement so if I do not get time to implement it myself, it will exist as a goal for the future.

README.md
```diff
@@ -297,12 +289,23 @@ firewall { '002 drop NEW external website packets with FIN/RST/ACK set and SYN u
   state     => 'NEW',
   action    => 'drop',
   proto     => 'tcp',
-  sport     => ['! http', '! 443'],
+  sport     => ['! http', '443'],
```
Ramesh7, line 292: Doesn't seems to be good to have inversion for first value only, user can make mistake by putting inversion on each values. Coz user is dump and expect them to over read the inversion guidelines.
I think the earlier design was good where user need to put inversion for all values. Open for thoughts.

david22swan: I had done this mostly for convenience and to mimic more closely how firewall actually behaves, it seemed wasteful to me to have a bunch of extra negations in the array.
I could implement code so that it accepts either behaviour.
Had also considered simply having them add an exclamation as an array variable, i.e. ` ['!', 'http', '443']`

LukasAud (dismissed):
Looking Good To Me. Hoping we will get a bunch more reviews. This deserves a lot of 👀

README.md
david22swan: Have updated the code to allow for either behaviour, the relevant array attributes must either have only the first value negated, or be negated in their entirety.

Ramesh7:
LGTM, can we get review from more experts?

david22swan:
@bastelfreak Could you take a look at this?

.sync.yml
```diff
@@ -6,6 +6,7 @@ appveyor.yml:
 Gemfile:
   optional:
     ":development":
+      - gem: 'puppet-resource_api'
```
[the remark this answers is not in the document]
david22swan: It relies partly on the version brought in by the puppet agent, so pinning here may cause conflicts unless we update it to match every time.....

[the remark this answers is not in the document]
jordanbreen28: Cool - forget what I said then 😄

spec/spec_helper_local.rb
```diff
@@ -29,15 +29,53 @@
...
 
-def with_debian_facts
+shared_context 'ArchLinux' do
   let :facts do
     {
       kernel: 'Linux',
       os: {
-        name: 'Debian',
-        release: { full: '8.0' },
+        name: 'ArchLinux',
```
bastelfreak, line 37: why do you mock the facts here for arch linux? facterdb has a factset for it

spec/unit/classes/firewall_linux_archlinux_spec.rb
```diff
@@ -3,23 +3,17 @@
 require 'spec_helper'
 
 describe 'firewall::linux::archlinux', type: :class do
-  let(:facts) do
```
bastelfreak, line 6: all of the changes here aren't related to the resource_api I think? Should they be pulled out as well?

.rubocop.yml
```diff
@@ -88,3 +88,6 @@ Style/Documentation:
   - spec/**/*
 Style/WordArray:
   EnforcedStyle: brackets
+Naming/MethodParameterName:
+  AllowedNames:
+    - 'is'
```
bastelfreak, line 93: is there `\n` missing in line 93?

README.md
```diff
@@ -170,15 +170,7 @@ resources { 'firewallchain':
 Internal chains can not be deleted. In order to avoid all the confusing
 Warning/Notice messages when using `purge => true`, like these ones:
```
bastelfreak, line 172: can you add a migration path to the README.md / document the breaking changes?

spec/spec_helper_local.rb
david22swan: updating to:
```
shared_context 'ArchLinux' do
  let(:facts) { on_supported_os['archlinux-x86_64'] }
end
```

spec/spec_helper_acceptance_local.rb
```diff
@@ -21,7 +21,7 @@ def ip6tables_flush_all_tables
 
 def install_iptables
   LitmusHelper.instance.run_shell('iptables -V')
-rescue
+rescue StandardError
```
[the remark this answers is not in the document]
david22swan: Ah, the changes here and in other similar situations are rubocop fixes that where corrected automatically when I was running my updated code through the process.
So while they're not all strictly related to the adoption of resource_api they are somewhat relevant under the theme of updating the module as a whole.
Or at least thats how I saw it.

spec/unit/classes/firewall_linux_archlinux_spec.rb
david22swan: Somewhat same as above, part of the update process was down as doing a full rubocop of the module to confirm the new code matched standards, and the autofixes brought in a large amount of fixes.

README.md
david22swan: Will do, have a blog post I can adapt so shouldn't take long.

spec/unit/classes/firewall_linux_archlinux_spec.rb
bastelfreak: mhm I understand that, but I would prefer it to have the rubocop changes in another PR. half of the diff is rubocop related.

README.md
david22swan: Have added

spec/spec_helper_local.rb
david22swan: Ok, this doesn't seem to work

david22swan:
CentOS 8 failures have appeared on nightly

spec/unit/classes/firewall_linux_archlinux_spec.rb
david22swan: You have a point, will se what I can do.
Was thinking of doing a full rebase and squish anyway, my commits kinda got away from me.

david22swan: Ok, that's all the unrelated rubocop fixes taken out and all the commits squished and cleaned up.
Will put up a second PR once this one gets merged.

spec/spec_helper_local.rb
david22swan: Including archlinux as I linked above does not seem to work.
Looking into facterdb it seems that archlinux is no longer included in the newest facts, i.e. 4.3 forward
https://github.com/voxpupuli/facterdb/tree/master/facts/4.3

README.md
```diff
@@ -542,3 +548,60 @@ And run the tests from the root of the source code:
 ```text
 bundle exec rake parallel_spec
 ```
+
+See also `.travis.yml` for information on running the acceptance and other tests.
```
bastelfreak, line 552: there is no travis.yml anymore.

README.md
```diff
@@ -542,3 +548,60 @@ And run the tests from the root of the source code:
...
+
+### Migration path to v7.0.0
+
+As of `v7.0.0` of this module a major rework has been done to adopt the [puppet-resource_api](https://github.com/puppetlabs/puppet-resource_api) into the module and use it style of code in place of the original form of Puppet Type and Providers. As part of this several breaking changes where made to the code that will need to be accounted for whenever you update to this new version of the module.
```
bastelfreak, line 556: looks god to me. Maybe one sentence why this change is made?

README.md
david22swan: Removed and replaced with a mention of the Github Actions

README.md
david22swan: Updated to include the primary reasoning

bastelfreak (approved):
I'm not a resource_api expert, but the changes look good to me.

LukasAud (approved):
LGTM 👍 

Dont forget to remove the Do Not Merge

david22swan:
@traylenator The functionality behind the attribute is still there, it just means renaming the attribute your setting to jump instead.
I felt it was wasteful to have two attributes for the same iptables value.
It may be inconvenient in the short term but in the long term it helps to simplify the code.

traylenator:
> @traylenator The functionality behind the attribute is still there, it just means renaming the attribute your setting to jump instead. I felt it was wasteful to have two attributes for the same iptables value. It may be inconvenient in the short term but in the long term it helps to simplify the code.

Thanks for the reply - will live status quo till we launch EL7 into oblivion (and drop this module).

Issue #1164, opened 2023-09-22
The jump parameter should not be converted to uppercase


sbraz:
## Describe the Bug
Hi,
I noticed a discrepancy between the behaviour of the `chain` and `jump` parameters. According to https://www.netfilter.org/documentation/HOWTO/packet-filtering-HOWTO-7.html, user-defined chains should use lowercase names.
> By convention, user-defined chains are lower-case to distinguish them (we'll describe how to create new user-defined chains below in [Operations on an Entire Chain](https://www.netfilter.org/documentation/HOWTO/packet-filtering-HOWTO-7.html#chain-ops)).

If I follow this convention, it works when passed to `chain` but `jump` forcefully converts my chain name to uppercase.

## Expected Behavior
The value of the `jump` parameter should not be forced to uppercase.

## Steps to Reproduce
Steps to reproduce the behaviour:
* Add directives to create a custom chain with a lowercase name and a rule which jumps to this chain.
* It fails because `returned 2: iptables v1.8.7 (nf_tables): Chain 'CUSTOM' does not exist`

## Environment
 - Version 7.0.2
 - Platform Debian 11

## Additional Context
This line should be removed:
https://github.com/puppetlabs/puppetlabs-firewall/blame/v7.0.2/lib/puppet/provider/firewall/firewall.rb#L867

david22swan:
This upcase was put in place due to the merge with `action`, which had allowed required the ACCEPT, REJECT and DROP values to be passed as lowercase.
The upcase was an attempt to keep the code as compliant as possible but had not considered the custom values.
Will clean it up now.

PR #1165, opened 2023-09-26
(GH-1164) Only common jump values should be enforced as upcase

david22swan:
Common jump values, depending on the OS, can be returned either as upcase or downcase values causing idempotentcy issues. This had previously been dealt with by simply enforcing upcase on all passed values. However this caused issues when passing chain's as the jump target and so the code has been update to explicitly upcase the common values only.

## Summary
Provide a detailed description of all the changes present in this pull request.

## Additional Context

## Related Issues (if any)

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.


gavindidrichsen (approved):
LGTM

PR #1206, opened 2024-04-04
Fix "creation" of empty built-in firewall chains

2fa:
## Summary
Fixes the issue when you are trying to create an in-built firewall chain in a table that was not been interacted with before. 

## Additional Context
Currently if you are trying to create in-built chain in an empty table (nat, for example) puppet will apply manifest without actually doing anything every time you run it. This is because firewall module relyes on `iptables-save` output that will not contain in-built chain if corresponding table not been interacted with yet. 

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.


david22swan (approved):
LGTM

Thanks for catching this.

Ramesh7 (approved):

PR #1207, opened 2024-04-04
Allow -_ in ipset name

### LINKED ISSUE #1187: Hyphen in the ipset's hash name breaks a firewall resource
## Describe the Bug
For example, we have the ipset hash named "wan-networks"

Applying a catalog generates the next error
```
Error: Failed to apply catalog: Parameter ipset failed on Firewall[200 allow tcp from wan-networks src in INPUT for eth0]: firewall.ipset expects a value of type Undef, Pattern[/^(?:!\s)?\w+\s(?:src|dst)(?:,src|,dst)?$/], or Array[Pattern[/^(?:!\s)?\w+\s(?:src|dst)(?:,src|,dst)?$/]], got String
```

## Expected Behavior
No error, as it was before 7.0.0

## Steps to Reproduce
1. Make a hash of addresses with the ipset containing hyphen in its name.
1. Try to apply some rule with that hash

## Environment
 - Version 7.0.2
 - Puppet: 8.4.0

2fa:
## Summary
Allow setting a `-` and `_` in ipset name

Should fix #1187 

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -879,7 +879,7 @@ Data type: `Optional[Enum['none', 'ipsec']]`
 
 ##### `ipset`
 
-Data type: `Optional[Variant[Pattern[/^(?:!\s)?\w+\s(?:src|dst)(?:,src|,dst)?$/], Array[Pattern[/^(?:!\s)?\w+\s(?:src|dst)(?:,src|,dst)?$/]]]]`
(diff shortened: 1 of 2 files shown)


2fa:
> @2fa FYI the "\w" already contains underscores, so as a workaround one can change - with _ in existing ipsets.
> 
> This fix would still be great if merged, as it used to work before puppetlabs-firewall 7.0.0 release.

Yeah, you are correct about `_`. I probably was fixing `-` and thought about to add underscore as a future-proof, but forgot that it's already included in `\w`.

I know that i can simple rename all ipsets to use underscores, but it worked before so for me it was better to just fix the issue.

Right now we're using the latest 6.X version, 7.0 has too many problems: it's very slow, for example. #1214

PR #1208, opened 2024-04-04
Single element ipset array should be in sync with its string equivalent

2fa:
## Summary
Currently if you have a single value `ipset` parameter it will always be trying to correct itself on every puppet run.

## Additional info
I tried to create proper acceptance test for this case but it requires installed `ipset` package with created ipsets for testing so i decided not to do it.
## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.


david22swan:
@2fa Closing this in favor if https://github.com/puppetlabs/puppetlabs-firewall/pull/1286

Thanks for raising this though and I have made sure that you are properly credited for the work you provided
Hope to see more from you in the future.

PR #1209, opened 2024-04-04
Fix table parsing with * in non-table names

2fa:
## Summary
If `iptables-save` contains `*` in a non-table names it will break table parsing: 

1. Current table rules after encountered `*` will be considered `absent` and will be forcefully added on every puppet run without duplicate checks (cuz puppet thinks that there are no more rules in that table)
2. Next table will be skipped completely. 
 
## Additional Context
We need to have more `iptables-save` parsing tests, there are more bugs at rule level as well (i will open PR tomorrow)

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.


david22swan:
@2fa Closing this in favour of https://github.com/puppetlabs/puppetlabs-firewall/pull/1285

Thank you for bringing this to our attention however and I have made sure to give you credit for the original work on the new PR.
Hope to see more fro you in the future

Issue #1182, opened 2023-11-30
No value is detected for nflog-prefix


cgrycki:
## Describe the Bug
When the module parses existing rules with the nflog-prefix property, it incorrectly identifies the value as an empty string. As a result, every time the puppet agent runs, it detects a change to nflog_prefix and updates the rule with the same value.

The issue seems to stem from an additional space between "--nflog-prefix" and the value in the output of iptables-save and ip6tables-save. The regex that parses this rule appears to be expecting a single space, but there are two spaces.
https://github.com/puppetlabs/puppetlabs-firewall/blob/3ff86aab2b9a7010ab136ebac7a274631c4a3a68/lib/puppet/provider/firewall/firewall.rb#L535

## Expected Behavior
The module identifies the correct value of the nflog-prefix property when it parses rules from the output of iptables-save and ip6tables-save.

## Steps to Reproduce
1. Use this module to manage a rule with the nflog_prefix property configured.
2. Observe from the output of the puppet agent that the rule is updated every time that the agent runs.

## Environment
 - puppetlabs-firewall 7.0.2
 - Puppet Enterprise 2021.7.6
 - Puppet Agent 7.27.0
 - Ubuntu 22.04

## Additional Context
I have also confirmed on Ubuntu 20.04 that the output of iptables-save has two spaces between --nflog-prefix and the value.

weastur:
+1

durist:
+1

Issue #1170, opened 2023-10-24
Firewall chains can't have "-A" in their names of the firewall rule regex gest confused / fails


minorOffense:
## Describe the Bug
When debugging the upgrade to the v7 of puppetlabs-firewall our rules stopped working. We were getting

```
```

After a lot of debugging, I found that the regex in `def self.rule_to_name` was getting confused with the `-A` in our firewall chain named `COLDFRONT-ACCESS`.

See the debug output I had added here:

```
Notice: firewall[006 Blocked IPs (tcp) from 50.63.2.0]: Creating: List command execute result "# Generated by iptables-save v1.8.8 (nf_tables) on Tue Oct 24 18:06:03 2023\n*filter\n:INPUT ACCEPT [40051:14954037]\n:FORWARD ACCEPT [0:0]\n:OUTPUT ACCEPT [270851:177330168]\n:COLDFRONT-ACCESS - [0:0]\n:DROPFORT-ACCESS - [0:0]\n:DROPFORT-DENY - [0:0]\n:DROPFORT-SMURF-PROTECTION - [0:0]\n:DROPFORT-SYN-FLOOD - [0:0]\n:DROPFORT-TCP-DDOS - [0:0]\n:DROPFORT-UNMANAGED - [0:0]\n:ZABBIX-AGENT - [0:0]\n-A INPUT -p icmp -j ACCEPT\nCOMMIT\n# Completed on Tue Oct 24 18:06:03 2023\n"
Notice: firewall[006 Blocked IPs (tcp) from 50.63.2.0]: Creating: Listing for table *filter
:INPUT ACCEPT [40051:14954037]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [270851:177330168]
:COLDFRONT-ACCESS - [0:0]
:DROPFORT-ACCESS - [0:0]
:DROPFORT-DENY - [0:0]
:DROPFORT-SMURF-PROTECTION - [0:0]
:DROPFORT-SYN-FLOOD - [0:0]
:DROPFORT-TCP-DDOS - [0:0]
:DROPFORT-UNMANAGED - [0:0]
:ZABBIX-AGENT - [0:0]
-A INPUT -p icmp -j ACCEPT
COMMIT
# Completed on Tue Oct 24 18:06:03 2023

Notice: firewall[006 Blocked IPs (tcp) from 50.63.2.0]: Creating: This is a Rule ["-ACCESS - [0:0]"]
[...]

PR #1210, opened 2024-04-05
(MODULE-11463): Fix rule parsing when iptables chains with '-A' in the name

2fa:
## Summary
This PR fixes a couple of issues, related to iptables-save output parsing:

1. Double quotes in the comment breaks rule parsing (part after the first `"` will be ignored)
2. Rule parser thinks that custom chains with -A in the name (`NOMAD-ADMIN`, for example) is a start of the rule
3. Single quotes in the comments generates invalid `iptables-save` output that will then be processed by the rule parser.
4. For some parameters `iptables-save` return multiple spaces between flag and actual value (`--nflog--prefix`)

## Additional Context
I've added additional spec test that will be helpful with `iptables-save` output parsing in the future.

## Related Issues (if any)
#1182 
#1170 

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.


chrisongthb:
Nice work! Works in our environment on Ubuntu 22.04 & 20.04 k8s nodes. Thanks!

span786:
Due to Rubocop failures in this PR, the commits from it were used in https://github.com/puppetlabs/puppetlabs-firewall/pull/1248 to address those issues.

malikparvez (approved):
Looks good...

Issue #1198, opened 2024-03-19
issue with match_mark regex


robertc99:
I getting a warning out of the firewall module 8.0.0) on a rule created by podman
The warning is

Warning: Provider returned data that does not match the Type Schema for `firewall[9031 netavark portfw masq mark]`
 Value type mismatch:
    * match_mark: 0x2000/0x2000 (expects an undef value or a match for Pattern[/^(?:!\s)?[a-fA-F0-9x]+$/], got '0x2000/0x2000')

I believe rules causing the problem are
-A NETAVARK-HOSTPORT-MASQ -m comment --comment "netavark portfw masq mark" -m mark --mark 0x2000/0x2000 -j MASQUERADE
-A NETAVARK-HOSTPORT-SETMARK -j MARK --set-xmark 0x2000/0x2000

robertc99:
Ive got a 2 character fix. I'll do a PR when I get a chance

PR #1211, opened 2024-04-05
Add ability to set match_mark|connmark as a mark/mask value

2fa:
## Summary
You should be able to set mark|connmark as a mark/mask. This PR add support of that.

## Related Issues (if any)
#1198 

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -364,10 +364,10 @@ Data type: `Optional[Integer]`
 
 ##### `connmark`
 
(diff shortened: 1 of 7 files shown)


absld:
Any updates on this?

absld:
Can confirm that the issue is solved with 4354dd8 from @2fa

span786:
We will be adding the changes mentioned in this PR in https://github.com/puppetlabs/puppetlabs-firewall/pull/1238.

PR #1212, opened 2024-04-05
Port ranges as string with hyphen as range indicator should work

2fa:
## Summary
If you set `dport|sport` values as a string with `-` as range indicator puppet will be applying manifest on every run.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/firewall/firewall.rb
+++ b/lib/puppet/provider/firewall/firewall.rb
@@ -443,8 +443,6 @@ class Puppet::Provider::Firewall::Firewall
         should = should.tr('-', ':') if ports.include?(property_name)
 
         is == should
-      else
-        return nil
       end
     when :string_hex
       # Compare the values with any whitespace removed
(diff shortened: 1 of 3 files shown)


shubhamshinde360:
When dport/sport is passed as a string "561-570" the insync? method always returns nil since we only process the ports [in case they are arrays ](https://github.com/puppetlabs/puppetlabs-firewall/blob/40d2d35c291071612798075fd8c9eada9d6da469/lib/puppet/provider/firewall/firewall.rb#L437)otherwise return nil (falsey).
This results in the puppet applying the change even when the desired port ranges are the same as actual port ranges.

With this change the string port ranges are handled properly and the insync? method returns true when the port ranges are equal.

shubhamshinde360:
Thanks @2fa,

Looks good to me.
Could you please rebase your branch with main before this is merged.

shubhamshinde360 (approved):

2fa:
@shubhamshinde360 sorry, i was quite busy yesterday and just got your message. Do i need to rebase the branch or all is good now?

2fa:
@shubhamshinde360 fixed the issue with rubocop but there are failed acceptance tests. I don't think that's on my side though.

shubhamshinde360 (approved):

PR #1218, opened 2024-04-30
Allowing ICMP types.

oliparcol:
## Summary

The regexp used to convert ICMP string codes to integer was updated in the v7 refactor and the new one doesn't allow anymore passing strings of the type `<icmp type>/<icmp code>` (e.g. `3/4`). This is used to generate rules of the type:
```
 pkts bytes target     prot opt in     out     source               destination         
    0     0 ACCEPT     1    --  *      *       0.0.0.0/0            0.0.0.0/0            state NEW icmptype 3 code 4 /* Allow ICMP fragmentation needed */
```

The function `icmp_name_to_number` returns `nil` when a string of the type `<icmp type>/<icmp code>` is used in the `icmp` property.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [X] Manually verified. (For example `puppet apply`)

### DIFF
--- a/lib/puppet_x/puppetlabs/firewall/utility.rb
+++ b/lib/puppet_x/puppetlabs/firewall/utility.rb
@@ -145,7 +145,7 @@ module PuppetX::Firewall # rubocop:disable Style/ClassAndModuleChildren
 
     # Translate the symbolic names for icmp packet types to integers
     def self.icmp_name_to_number(value_icmp, protocol)
-      if value_icmp.to_s.match?(%r{^\d+$})
+      if value_icmp.to_s.match?(%r{^(\d+|\d+\/\d+)$})
         value_icmp.to_s
       elsif ['IPv4', 'iptables'].include?(protocol)
         # https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml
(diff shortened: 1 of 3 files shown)


lib/puppet_x/puppetlabs/firewall/utility.rb
```diff
@@ -145,7 +143,7 @@ def self.host_to_mask(value, proto)
 
     # Translate the symbolic names for icmp packet types to integers
     def self.icmp_name_to_number(value_icmp, protocol)
-      if value_icmp.to_s.match?(%r{^\d+$})
+      if value_icmp.to_s.match?(%r{^(\d+|\d+\/\d+)$})
```
david22swan, line 148:
```suggestion
      if value_icmp.to_s.match?(%r{^\d+(\/\d+)?$})
```
The regex you've suggested is slightly redundant, this may be a cleaner change

spec/unit/puppet_x/puppetlabs/firewall/utility_spec.rb
```diff
@@ -170,6 +150,7 @@
       it { expect(utility.icmp_name_to_number('timestamp-reply', proto)).to eql '14' }
       it { expect(utility.icmp_name_to_number('address-mask-request', proto)).to eql '17' }
       it { expect(utility.icmp_name_to_number('address-mask-reply', proto)).to eql '18' }
+      it { expect(utility.icmp_name_to_number('3/4', proto)).to eql '3/4' }
```
david22swan, line 173: This is a good test, but it would be nice to test if the code handles multi digit values, i.e. change the `4` to `40`?

david22swan (commented):
Some small changes would be nice, bu overall looks to be in a good condition

david22swan:
@oliparcol Closing this in favour of https://github.com/puppetlabs/puppetlabs-firewall/pull/1290

Thank you for putting in the work however and we hope to see more from you in the future.

PR #1234, opened 2024-10-04
nfmask, ctmask, restore_mark support + firewallchain loop fix

hb0nes:
## Summary
We have been using our custom version of puppetlabs-firewall for some time because we need the nfmask, ctmask and restore_mark flags for policy based routing.

While upgrading to the latest puppetlabs-firewall, I ran into an infinite loop, creating chains all of a sudden.
As per the docs, I've written this Puppet code:
```
  # Purge all chains not defined below
  resources { 'firewallchain':
    purge  => true,
  }

  # Purge all unmanaged rules from these (internal) chains
  # These have default policy DROP
  firewallchain {
    [
      'INPUT:filter:IPv4',
      'INPUT:filter:IPv6',
      'OUTPUT:filter:IPv4',
      'OUTPUT:filter:IPv6',
      'FORWARD:filter:IPv4',
      'FORWARD:filter:IPv6']:
        ensure => present,
        policy => drop,
        purge  => true,
  }

  # The internal chains need to be defined or Puppet will try to remove them.
  # We can also purge their contents by defining them here with purge => true.
  firewallchain {
    [
      # IPv4 mangle
      'PREROUTING:mangle:IPv4',
      'INPUT:mangle:IPv4',
      'FORWARD:mangle:IPv4',
      'OUTPUT:mangle:IPv4',
      'POSTROUTING:mangle:IPv4',
      # IPv6 mangle
      'PREROUTING:mangle:IPv6',
      'INPUT:mangle:IPv6',
...and so forth
```
Because iptables-save doesn't always show chains like *nat or *mangle if they haven't been interacted with, the `get` function will not return those and puppetlabs-firewall will try to create the chains. However, this won't do anything and the infinite loop starts.

I have 'solved' it by creating those chains manually for now and only adding them if the already existing code didn't find them.

I have verified that these changes solve my issue and that nfmask, ctmask and restore_mark work.

Because this actually is an issue that I imagine others will run into, I decided to create the PR, regardless of code quality, lack of context and knowledge.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [X] Manually verified. (For example `puppet apply`)

### DIFF
--- a/lib/puppet/provider/firewallchain/firewallchain.rb
+++ b/lib/puppet/provider/firewallchain/firewallchain.rb
@@ -47,6 +47,40 @@ class Puppet::Provider::Firewallchain::Firewallchain
   def get(_context)
     # Create empty return array
     chains = []
+    # Built-in chains are not always visible in iptables-save, but they are still there.
+    # Add them to the chains if they aren't found by iptables-save.
+    built_in_chains = []
+    [
+      # filter table
+      "FORWARD:filter",
+      "INPUT:filter",
+      "OUTPUT:filter",
+      # mangle table
+      "FORWARD:mangle",
+      "INPUT:mangle",
+      "OUTPUT:mangle",
+      "POSTROUTING:mangle",
+      "PREROUTING:mangle",
+      # nat table
+      "INPUT:nat",
+      "OUTPUT:nat",
+      "POSTROUTING:nat",
+      "PREROUTING:nat",
+      # raw table
+      "OUTPUT:raw",
+      "PREROUTING:raw"
+    ].each do | chain_table |
+      ['IPv4', 'IPv6'].each do |protocol|
+          name = "#{chain_table}:#{protocol}"
+          chain_hash = {
+            name: name,
+            purge: false,
+            ignore_foreign: false,
+            ensure: 'present'
+          }
+          built_in_chains << chain_hash
+      end
+    end
     # Scan String to retrieve all Chains and Policies
     ['IPv4', 'IPv6'].each do |protocol|
       # Go through each supported table and retrieve its chains if it exists.
@@ -66,7 +100,13 @@ class Puppet::Provider::Firewallchain::Firewallchain
         end
       end
     end
-    # Return array
+    # Add built_in_chains to chains only if they haven't already been found
+    built_in_chains.each do |built_in_chain|
+      unless chains.any? { |chain| chain[:name] == built_in_chain[:name] }
+        chains << built_in_chain
+      end
+    end
+
     chains
   end
 
--- a/lib/puppet/type/firewall.rb
+++ b/lib/puppet/type/firewall.rb
@@ -593,6 +593,18 @@ Puppet::ResourceApi.register_type(
         In order to maintain compatibility it is also possible to negate all values given in the array to achieve the same behaviour.
       DESC
     },
+    ctmask: {
+      type: 'Optional[String]',
+      desc: <<-DESC
+      ctmask
+      DESC
+    },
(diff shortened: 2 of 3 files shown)


lib/puppet/type/firewall.rb
```diff
@@ -593,6 +593,18 @@
         In order to maintain compatibility it is also possible to negate all values given in the array to achieve the same behaviour.
       DESC
     },
+    ctmask: {
+      type: 'Optional[String]',
+      desc: <<-DESC
+      ctmask
```
david22swan, line 599: A more thorough description would be needed here, to properly explain this attribute and its connection to the other two.
This applies to nfmask and restore_mark as well.

lib/puppet/type/firewall.rb
```diff
@@ -593,6 +593,18 @@
         In order to maintain compatibility it is also possible to negate all values given in the array to achieve the same behaviour.
       DESC
     },
+    ctmask: {
+      type: 'Optional[String]',
```
david22swan, line 597: Rather than simply string, a pattern similar to what `set_mark` uses would be preferred.
This applies to nfmask as well

It would also be a good idea to update the code so that they are treated the same as `set_mark` by the [insync?](https://github.com/hb0nes/puppetlabs-firewall/blob/bcd8e7a70f1d9094c9dcf619ac3e7e58921a5557/lib/puppet/provider/firewall/firewall.rb#L400) and the [process_input](https://github.com/hb0nes/puppetlabs-firewall/blob/bcd8e7a70f1d9094c9dcf619ac3e7e58921a5557/lib/puppet/provider/firewall/firewall.rb#L898) methods to ensure no issues.

Finally [validate_input](https://github.com/hb0nes/puppetlabs-firewall/blob/bcd8e7a70f1d9094c9dcf619ac3e7e58921a5557/lib/puppet/provider/firewall/firewall.rb#L743) should be updated to raise an error if either `ct_mask` or `nfmask` is set without `restore_mark`, as they require it in order to function correctly.

lib/puppet/provider/firewallchain/firewallchain.rb
```diff
@@ -7,52 +7,87 @@ class Puppet::Provider::Firewallchain::Firewallchain
...
+      "OUTPUT:nat",
+      "POSTROUTING:nat",
+      "PREROUTING:nat",
+      # raw table
+      "OUTPUT:raw",
+      "PREROUTING:raw"
+    ].each do | chain_table |
+      ['IPv4', 'IPv6'].each do |protocol|
+          name = "#{chain_table}:#{protocol}"
+          chain_hash = {
```
david22swan, line 75: A different fix for this issue has already been applied, `iptables-save` is now ran with the `-t` flag which will explicitly force the tables to initialise if it exists, regardless of whether or not it has been touched.

david22swan (changes requested):
Aside from the changes listed below we would also require some **Unit** tests to be added in order to ensure the functionality of your new values.

Apologies on the wait for this review, but if you can respond I hope to get your new functionality merged in promptly.

lib/puppet/provider/firewallchain/firewallchain.rb
david22swan: In addition, when `Create` is called against a built in chain, it runs `-L` instead of `-N`, initialising it rather than creating it.

david22swan:
@hb0nes Closing this in favour of https://github.com/puppetlabs/puppetlabs-firewall/pull/1291

Thank you for putting in the work on this and we hope to see more from you in the future.

PR #1237, opened 2024-10-21
(CAT-2088): Fix for not rejecting valid iptables

span786:
Fix for not rejecting valid iptables:
1. Allow -_ in ipset name


malikparvez:
This will fix it #1207

malikparvez (approved):

PR #1248, opened 2024-11-21
(MODULE-11463): Fix rule parsing when iptables chains with '-A' in the name

span786:
## Summary

Commits cherry-picked from https://github.com/puppetlabs/puppetlabs-firewall/pull/1210

This PR fixes a couple of issues, related to iptables-save output parsing:

Double quotes in the comment breaks rule parsing (part after the first " will be ignored)
Rule parser thinks that custom chains with -A in the name (NOMAD-ADMIN, for example) is a start of the rule
Single quotes in the comments generates invalid iptables-save output that will then be processed by the rule parser.
For some parameters iptables-save return multiple spaces between flag and actual value (--nflog--prefix)

## Additional Context
I've added additional spec test that will be helpful with iptables-save output parsing in the future.

## Related Issues (if any)
https://github.com/puppetlabs/puppetlabs-firewall/issues/1182
https://github.com/puppetlabs/puppetlabs-firewall/issues/1170

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


span786:
This will be merged as a part of https://github.com/puppetlabs/puppetlabs-firewall/pull/1210

PR #1251, opened 2024-12-20
Query gid to prevent errors with missing users with the same id (fixes #1229)

### LINKED ISSUE #1229: gid match fails if no user by the same name exists.
## Describe the Bug
Recently updated from 6.0.0 to 8.0.1. Now when using the gid match it looks for a user by the same name as the group. If no such user exists it fails with the error "can't find user for group name".

## Expected Behavior
Should do a gid lookup for the group name and not need or care if a user by the same name exists.

## Steps to Reproduce
1. Create a new group (called power-users in below example)
2. Make sure no user called "power-users" exists
3. Run example code below

## Environment
 - Version 8.0.1
 - Puppet 7.31.0
 - Platform Debian 12.6

## Additional Context
Example puppet code:
```puppet
firewall {
  default:
    chain => 'OUTPUT',
    jump  => 'accept',
    proto => 'tcp',
    dport => 123;
  '013 Accept for power-users':
    gid => 'power-users';
  # Removed other rules for brevity
}
```
Error:
```
Info: Unknown failure using insync_values? on type: Firewall[013 Accept for power-users] / property: gid to compare values ["power-users"] and 1420                                                               
Error: /Stage[main]/Profile::Lydia::Outbound_mail_firewall/Firewall[013 Accept for power-users]/gid: change from '1420' to 'power-users' failed: can't find user for power-users                                  
Error: can't find user for power-users
```

cmusik:
## Summary
If a firewall rule is assigned to a group and the same user does not exist, then we get an error:

```
Error: /Stage[main]/Main/Firewall[801 - gid testgroup]/gid: change from '1234' to 'testgroup' failed: can't find user for testgroup
```

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ x] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/firewall/firewall.rb
+++ b/lib/puppet/provider/firewall/firewall.rb
@@ -376,9 +376,16 @@ class Puppet::Provider::Firewall::Firewall
       end
 
       # If 'is' or 'should' contain anything other than digits or digit range,
-      # we assume that we have to do a lookup to convert to UID
-      is = Etc.getpwnam(is).uid unless is[%r{[0-9]+(-[0-9]+)?}] == is
-      should = Etc.getpwnam(should).uid unless should[%r{[0-9]+(-[0-9]+)?}] == should
+      # we assume that we have to do a lookup to convert to UID or GID
+      if property_name == :uid
+        is = Etc.getpwnam(is).uid unless is[%r{[0-9]+(-[0-9]+)?}] == is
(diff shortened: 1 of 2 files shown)


amitkarsale:
@cmusik, Thanks for the PR, can you please check for the test failures.

cmusik:
I have checked the logs, but the errors are during puppet agent setup phase. I think that we have an infrastructure problem here and I can't do anything about it.

PR #1261, opened 2025-04-29
Fix managing build-in chain on iptables nft

### LINKED ISSUE #1217: Non idempotent logs for empty firewall chains
## Describe the Bug

When applying on a server without any iptables rule the following puppet code with the resource firewallchain declared without any rule:
```
firewallchain { "FORWARD:mangle:IPv4":
    ensure => present,
    purge  => true,
}
```

The following output is always emitted:
```
Notice: /Stage[main]/Base::Firewall/Firewallchain[FORWARD:mangle:IPv4]/ensure: defined 'ensure' as 'present'
Notice: firewallchain[FORWARD:mangle:IPv4]: Updating: Finished in 0.000061 seconds
```

## Expected Behavior

I would expect no output to be emitted.

## Environment
 - Firewall 8.0.1
 - Ubuntu 24.04

## Additional Context

I believe that the issue comes from the fact that `iptables-save` doesn't show empty tables. The code is therefore not able to distinguish an existing empty table from a non-existing one. Specifying the table with the `-t` option (e.g. `iptables-save -t <table>` does display the empty table).

waipeng:
`iptables-save` does not show any build-in chains when they have not been interacted with. This leads to the chain being (re)-created.

#1206 attempted to fix this, by listing the chain instead of creating it, but this method only seems to work on non-nft iptables (https://github.com/puppetlabs/puppetlabs-firewall/issues/1217#issuecomment-2130211647).

Fix this for nft version of iptables by setting the policy of the chain instead of listing it. This seems to work for both nft and non-nft version of iptables (tested on Ubuntu 20.04 and 24.04).

Fixes #1217

## Summary
Provide a detailed description of all the changes present in this pull request.

## Additional Context

## Related Issues (if any)

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/firewallchain/firewallchain.rb
+++ b/lib/puppet/provider/firewallchain/firewallchain.rb
@@ -34,10 +34,6 @@ class Puppet::Provider::Firewallchain::Firewallchain
   $chain_delete_command = '-X'
   # Command to set chain policy, works on inbuilt chains only
   $chain_policy_command = '-P'
-  # Command to list specific table so it will generate necessary output for iptables-save
-  # The retrieval of in-built chains may get confused by `iptables-save` tendency to not return table information
-  # for tables that have not yet been interacted with.
-  $table_list_command = '-L'
   # Check if the given chain name references a built in one
   $built_in_regex = %r{^(?:INPUT|OUTPUT|FORWARD|PREROUTING|POSTROUTING)$}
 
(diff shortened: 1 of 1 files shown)


david22swan:
Hey, closing this as an alternative fix has been merged: https://github.com/puppetlabs/puppetlabs-firewall/pull/1263
Thanks for putting in the work to fix however.

PR #1263, opened 2025-06-03
(MODULES-11476) Fix non-idempotency of firewall table creation

shubhamshinde360:
## Summary
Fix non-idempotency of firewall table creation

## Additional Context
The 'iptables-save' command behaves differently on rpm and deb based platforms.
rpm based: If an iptable has been interacted with, it will show up in the command even if it is empty.
deb based: If an iptable has been interacted with, it may not show up in the command if it is empty. If we add and then delete a rule, it shows up. But if we create the table using "iptables -N" it doesn't.

Specific tables always show up with "iptables -t #{table_name}".
Since, the tables are always already created and the module supports a specific set of them. This commit tries to check if they exist one by one using the "iptables -t" command.

Also, since the tables are being checked like this, all tables show up in the output, so adjusted the rpsec according to that.

Root cause: This occurs because iptables do not show up in the 'iptables' command when they are empty in deb based distributions.

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/firewallchain/firewallchain.rb
+++ b/lib/puppet/provider/firewallchain/firewallchain.rb
@@ -15,8 +15,8 @@ class Puppet::Provider::Firewallchain::Firewallchain
   }
   # Regex used to divide output of$list_command between tables
   $table_regex = %r{(\*(?:nat|mangle|filter|raw|rawpost|broute|security)[^*]+)}
-  # Regex used to retrieve table name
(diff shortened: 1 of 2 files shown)


david22swan (approved):
LGTM

Smart change, wish I'd thought of it

david22swan (approved):
LGTM

PR #1267, opened 2025-08-05
(MODULES-11580) Do not combine stderr and stdout when scanning rules

shubhamshinde360:
## Summary
 - We get the output of 'ip(6)tables-save' and parse the contents to get the iptables rules.
 - When stderr is combined with stdout, the output of the command can be unexpected causing issues when parsing.
 - This commit disables the combination of stdout and stderr for this reason.
 - Setting 'failonfail' makes sure that any error returned by 'ip(6)tables-save' command is reported as failure to the user without attempting to parse its output.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


amitkarsale (approved):
LGTM

PR #1284, opened 2026-06-08
(bugfix) Fix log_level idempotency when explicitly setting the iptables default value

david22swan:
## Summary

iptables-save omits --log-level when the kernel default (4/warn) is in use. Treat a nil 'is' value as 4 in the log_level insync? comparison so that explicitly setting log_level to 4, 'warn', or 'warning' does not cause spurious changes on every Puppet run.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.


gavindidrichsen (approved):
LGTM

PR #1286, opened 2026-06-09
(bugfix) Fix ipset idempotency: single-element array not in sync with String equivalent

david22swan:
## Summary

Fixes an idempotency bug where a single-element `ipset` array (e.g. `ipset => ['setname src']`) was never considered in sync with the system state. The provider reads a single matched set back from `iptables-save` as a plain String, but the manifest value is an Array, causing `insync?` to fall through to the default comparison and report perpetual drift. The fix adds a dedicated `:ipset` case that normalises both sides to Array before comparing.

Rebase of https://github.com/puppetlabs/puppetlabs-firewall/pull/1208 with added tests

## Checklist

- [x] 🟢 Spec tests
- [x] 🟢 Acceptance tests
- [x] Manually verified

### DIFF
--- a/lib/puppet/provider/firewall/firewall.rb
+++ b/lib/puppet/provider/firewall/firewall.rb
@@ -450,6 +450,12 @@ class Puppet::Provider::Firewall::Firewall
       should = should_hash[property_name].to_s.gsub(%r{\s+}, '')
 
       is == should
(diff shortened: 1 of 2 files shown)


LukasAud (approved):
LGTM

PR #1288, opened 2026-06-09
(MODULES-11840) Allow puppetlabs/stdlib 10.x

imaqsood:
Part of MODULES-11840. Metadata-only change: stdlib dependency cap `< 10.0.0` -> `< 11.0.0` so this module installs alongside stdlib 10.x. No code change.


david22swan (approved):
LGTM

Planning a release of Firewall in the immeadiate future, will leave this till after

PR #1296, opened 2026-06-11
(#1254) Remove iptables-services from EL9+ package defaults

### LINKED ISSUE #1254: Remove requirement for iptables-services on EL9 generation systems
## Use Case
With the conversion of `iptables` based commands to use `nftables` under the hood, the package `iptables-services` is no longer being built and distributed along side their `nft` equivalents (i.e. `iptables-nft-services`). As such, when the `iptables` family of packages is updated, there is a delay in the matching `iptables-services` packages from EPEL being distributed.

As `iptables-nft-services` actually provides everything that this module requires, it should be safe to remove the reliance upon this EPEL package.

Remove the reliance upon EPEL for the `iptables-services` package for EL9 (and related FC) OS selections

As the reliance upon `iptables-services` is now reliant upon the 3rd group of EPEL making keeping packages up-to-date a mess, and the existing `iptables-nft-services` provides all the necessary source material, I felt no other alternative is worth further consideration.

## Additional Context
A simple test on EL9 instances with this diff does everything we need from the module.
```diff
diff --git a/manifests/params.pp b/manifests/params.pp
index 5fbe898..b62e9fe 100644
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -32,7 +32,7 @@ class firewall::params {
           if versioncmp($facts['os']['release']['full'], '9') >= 0 {
             $service_name = ['nftables','iptables']
             $service_name_v6 = 'ip6tables'
-            $package_name = ['iptables-services', 'nftables', 'iptables-nft-services']
+            $package_name = ['nftables', 'iptables-nft-services']
             $iptables_name = 'iptables-nft'
             $sysconfig_manage = false
             $firewalld_manage = true
```

david22swan:
## Summary

- Removes `iptables-services` from the default package list for EL9+ systems (RHEL, CentOS, AlmaLinux)
- Replaces it with `iptables-nft-services`, which is shipped by the distribution and provides the same functionality
- EL8 and earlier are unaffected; their `iptables-services` default is preserved

## Additional Context

On EL9+, `iptables` is implemented via nftables under the hood. The `iptables-services` package from EPEL belongs to the `iptables-legacy-*` stack and conflicts with `iptables-libs` updates, causing broken package state on systems that pull it in — even when Puppet wasn't responsible for installing it ([RHEL Bug 2343682](https://bugzilla.redhat.com/show_bug.cgi?id=2343682)).

`iptables-nft-services` is the correct EL9 equivalent: it is part of the official distribution and provides the service units needed by this module.

## Related Issues

Fixes #1254

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -32,7 +32,7 @@ class firewall::params {
           if versioncmp($facts['os']['release']['full'], '9') >= 0 {
             $service_name = ['nftables','iptables']
             $service_name_v6 = 'ip6tables'
-            $package_name = ['iptables-services', 'nftables', 'iptables-nft-services']
+            $package_name = ['iptables-nft-services', 'nftables']
             $iptables_name = 'iptables-nft'
             $sysconfig_manage = false
             $firewalld_manage = true
--- a/spec/unit/classes/firewall_linux_redhat_spec.rb
+++ b/spec/unit/classes/firewall_linux_redhat_spec.rb
@@ -37,6 +37,7 @@ describe 'firewall::linux::redhat', type: :class do
   ['RedHat', 'CentOS', 'Fedora', 'AlmaLinux'].each do |os|
     releases = ((os == 'Fedora') ? ['36'] : ['7.0.1406'])
     nftablesreleases = ((os == 'Fedora') ? [] : ['8.0'])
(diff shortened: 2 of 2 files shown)


gavindidrichsen (approved):
LGTM

PR #1309, opened 2026-08-03
(#1214) Cache iptables-save reads for the duration of a catalog run

### LINKED ISSUE #1214: Performance degradation in resource_api version
## Describe the Bug
After testing latest version of firewall module i've noticed that puppet applies catalog significantly longer compared to the previous version.

After a bit of strace it looks like new version makes several `iptables-save` calls every time it applies each rule.

```
Old version
~# grep -c "execve(\"/usr/sbin/iptables-save" strace1.out
60

New version
~# grep -c "execve(\"/usr/sbin/iptables-save" strace2.out
296
```

For some machines catalog application time increased more than 2x (from 50 to 120 seconds, for example).

## Expected Behavior
Firewall resource should not make multiple `iptables-save` calls for each rule.

## Environment
 - Version [3.1.0 and 8.0.1]
 - Platform [Ubuntu 20.04]

## Additional Context
It is related to #1100

AlexanderSalmin:
## Summary

Fixes the performance degradation reported in #1214, where catalog application time grew from seconds to many minutes after the Resource API migration.

The Resource API calls a provider's `get` once for every resource of that type in the catalog (`refresh_current_state` performs a full `get`, then `find`s the single resource being evaluated). For this module that means, per catalog run, even when nothing has changed:

- per `firewall` resource: one `iptables-save` + one `ip6tables-save` execution, plus a full regex parse of the entire ruleset (~150 attribute patterns per rule);
- per `firewallchain` resource: one `iptables-save -t <table>` execution for each of the 7 supported tables x 2 protocols.

A catalog with N rules on a host with M total rules therefore does O(N) command executions and O(N x M) parse work. The original report's "~296 iptables-save calls for ~60 rules" and the later reports of hour-long runs on hosts with ~1100 unmanaged rules follow directly from this.

This PR adds `PuppetX::Firewall::Cache`, a read cache for list-command output and its parsed form, shared by both providers and scoped to a single catalog application:

- Entries are keyed to the object identity of Puppet's `:current_environment` context binding. The agent binds a fresh `Puppet::Node::Environment` for every catalog run (`Configurer#push_current_environment_and_loaders`) and pops it when the run finishes, so a new run never observes a previous run's snapshot - external drift is still picked up on every run.
- Any `create`/`update`/`delete` of a rule or chain invalidates the whole cache, so the next read re-executes the underlying command and observes the change (the cache is shared between the two providers precisely so that chain flushes invalidate rule reads).
- When no `:current_environment` binding is available, caching is disabled and every read executes the underlying command, i.e. previous behaviour.

`get_rules` returns per-caller copies of cached entries (the hashes and their array values), so callers cannot alter the snapshot. This restores the read semantics of the legacy (pre-7.x) type/provider, whose `prefetch` also read the ruleset once per transaction.

Results, unchanged catalog, firewall resource evaluation only (synthetic benchmark):

| Scenario | v8.5.0 | this PR |
|---|---|---|
| 100 rules managed, 500 unmanaged | 40.5 s, 200 execs | 0.5 s, 2 execs |
| 100 rules managed, 4000 unmanaged | ~270 s, 200 execs | 3.9 s, 2 execs |

Validation on real infrastructure: a VPN gateway with 259 managed firewall resources and 274 live rules applies its full catalog in ~7 s under `--noop`, with a debug run confirming exactly one `iptables-save` and one `ip6tables-save` execution for the entire agent run.

## Related ticket / issue

- Ticket / issue: #1214

## Required checklist

- [x] Linked the driving ticket or issue above where one exists (a `MODULES-*` ticket via the structured spec template, or a GitHub issue).
- [x] Coverage gate is green (or the baseline is preserved -- coverage was not reduced). New unit specs cover the cache class and both providers' caching/invalidation behaviour; full suite passes (1314 examples).
- [ ] Platform matrix is green -- deferring to this PR's CI checks.
- [x] `Co-Authored-By: Claude` trailer is present on the commits if this PR was AI-assisted. This PR was AI-assisted (Claude); the trailer is on the commit.

## Review policy (outcome-based)

This PR is (check one):

- [ ] Auto-merge
- [ ] Single reviewer
- [x] Full review (2+)

## Additional context

**Regression analysis performed:**

1. *Stale reads after our own writes* - every mutation path in both providers invalidates the cache (chain flush/delete included). Covered by unit specs and an end-to-end drift test.
2. *Stale reads across runs* - the cache keys on the identity of the per-run environment binding; a new agent run, apply invocation, or `puppet resource` process never reuses a snapshot. Where the binding is absent (unusual embeddings), caching disables itself entirely.
3. *Mutation of shared parsed data* - audited the module and puppet-resource_api for in-place mutation of "is" values (`gsub!`, `map!`, `sort!`, index assignment); none apply to cached objects. Defensively, `get_rules` copies hashes and array values per caller anyway.
4. *Mid-run external mutations* (an `exec` running iptables, a service restart reloading saved rules between firewall resources) are no longer observed within the same run. This is identical to the legacy provider's prefetch-once semantics and self-corrects on the next run.
5. *Failure handling* - a failed `iptables-save` is not cached (the exception propagates; the next read retries).
6. *Unmanaged-rule synthetic names* (`9xxx <sha>`) - parse results are cached per call shape, so counter-derived names are identical to previous behaviour.

End-to-end tested with real iptables (nf_tables 1.8.11) and Puppet 8.10: chain + rule catalogs apply correctly, second run fully idempotent with each list command executed exactly once, externally deleted rules re-created in the correct position on the next run, `firewallchain` purge deletes unmanaged rules, `puppet resource firewall` works.

`spec_helper_local.rb` drops the cache before each example because the spec harness, unlike real runs, keeps one `:current_environment` binding across examples.

**Possible follow-ups (out of scope here):**

- Updating the cached snapshot in place after writes, so bootstrap runs creating many rules also avoid re-reads (currently each write invalidates, matching previous `insert_order` behaviour).
- Precompiling the ~150 attribute regexes in `rule_to_hash`/`rule_to_name`, which are currently rebuilt with `Regexp.new` for every rule parsed - this would speed up the one remaining full parse per run.
- Per-transaction `get` caching in puppet-resource_api itself, which would address this class of problem for all resource_api providers.

### DIFF
--- a/lib/puppet/provider/firewall/firewall.rb
+++ b/lib/puppet/provider/firewall/firewall.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative '../../../puppet_x/puppetlabs/firewall/utility'
+require_relative '../../../puppet_x/puppetlabs/firewall/cache'
 
 # Implementation for the iptables type using the Resource API.
 class Puppet::Provider::Firewall::Firewall
@@ -302,6 +303,7 @@ class Puppet::Provider::Firewall::Firewall
     position = Puppet::Provider::Firewall::Firewall.insert_order(context, name, should[:chain], should[:table], should[:protocol])
     arguments = Puppet::Provider::Firewall::Firewall.hash_to_rule(context, name, should)
     Puppet::Provider.execute([$fw_base_command[should[:protocol]], should[:table], $fw_rule_create_command, should[:chain], position, arguments].join(' '))
+    PuppetX::Firewall::Cache.invalidate
     PuppetX::Firewall::Utility.persist_iptables(context, name, should[:protocol])
   end
 
@@ -310,6 +312,7 @@ class Puppet::Provider::Firewall::Firewall
     position = Puppet::Provider::Firewall::Firewall.insert_order(context, name, should[:chain], should[:table], should[:protocol])
     arguments = Puppet::Provider::Firewall::Firewall.hash_to_rule(context, name, should)
     Puppet::Provider.execute([$fw_base_command[should[:protocol]], should[:table], $fw_rule_update_command, should[:chain], position, arguments].join(' '))
+    PuppetX::Firewall::Cache.invalidate
     PuppetX::Firewall::Utility.persist_iptables(context, name, should[:protocol])
   end
 
@@ -319,6 +322,7 @@ class Puppet::Provider::Firewall::Firewall
     # We do this to ensure accuracy when removing non-standard (i.e. uncommented) rules via the firewallchain purge function
     arguments = is[:line].gsub(%r{^-A}, $fw_rule_delete_command)
     Puppet::Provider.execute([$fw_base_command[is[:protocol]], is[:table], arguments].join(' '))
+    PuppetX::Firewall::Cache.invalidate
     PuppetX::Firewall::Utility.persist_iptables(context, name, is[:protocol])
   end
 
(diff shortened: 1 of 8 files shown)


jst-cyr:
CI test run created: https://github.com/puppetlabs/puppetlabs-firewall/pull/1313

jst-cyr:
**Acceptance Tests:** https://github.com/puppetlabs/puppetlabs-firewall/actions/runs/32860389450?pr=1313

All green except for SLES 12 which is a known issue, same provisioning issue happening in main with nightly runs. This PR does not seem to introduce any new issues.

From a testing perspective, this is looking good. Next step required is an engineering review when the team is available.

PR #1316, opened 2026-09-03
Fix perpetual corrective changes for proto with iptables >= 1.8.11

Thermikbomber:
## Summary

- iptables >= 1.8.11 (Debian 13+) deliberately removed `getprotobynumber()` from `iptables-save`, so rare protocols (vrrp, esp, ah, ospf, gre, …) are now emitted as their numeric IANA IDs instead of resolved names. This caused Puppet to report a corrective change on every run when the catalog used the symbolic name (e.g. `proto => 'vrrp'`) but `iptables-save` returned the number (`112`).
- Add `FALLBACK_PROTO_TABLE` / `PROTO_NAME_TO_NUMBER` constants to `Utility`, populated at load time from `/etc/protocols` with a comprehensive IANA fallback table for environments without that file (e.g. minimal containers).
- Replace the bare `else raise ArgumentError` in `proto_number_to_name` with a graceful reverse lookup so unknown protocol numbers are returned unchanged instead of crashing.
- Add `proto_name_to_number` helper method.
- Add `when :proto` case to `insync?` that normalises both `is` and `should` to their IANA protocol number before comparing, so `'112'` and `'vrrp'` are treated as equal regardless of which form `iptables-save` emits.

## Test plan

- [x] `rspec spec/unit/puppet_x/puppetlabs/firewall/utility_spec.rb` — 131 examples, 0 failures
- [x] `rspec spec/unit/puppet/provider/firewall/firewall_public_spec.rb` — 116 examples, 0 failures (includes 10 new proto `insync?` cases covering name↔number equivalence for vrrp, esp, ah, ospf, gre)
- [x] Deploy on a Debian 13 node with iptables 1.8.11 and verify no corrective change for `proto => 'vrrp'` rules

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/puppet/provider/firewall/firewall.rb
+++ b/lib/puppet/provider/firewall/firewall.rb
@@ -390,6 +390,15 @@ class Puppet::Provider::Firewall::Firewall
       end
 
       "#{is_negate}#{is}" == "#{should_negate}#{should}"
+    when :proto
+      # iptables >= 1.8.11 removed getprotobynumber() from iptables-save, so
+      # rare protocols (vrrp, esp, ah, ospf, gre, …) are now emitted as their
+      # numeric IANA IDs instead of resolved names.  Normalise both sides to
+      # their protocol number before comparing so that e.g. '112' and 'vrrp'
+      # are treated as equal.
+      is     = PuppetX::Firewall::Utility.proto_name_to_number(is_hash[property_name].to_s)
+      should = PuppetX::Firewall::Utility.proto_name_to_number(should_hash[property_name].to_s)
+      is == should
     when :mac_source, :jump
       # Value of mac_source/jump may be downcased or upcased when returned depending on the OS
       is_hash[property_name].casecmp(should_hash[property_name]).zero?
--- a/lib/puppet_x/puppetlabs/firewall/utility.rb
+++ b/lib/puppet_x/puppetlabs/firewall/utility.rb
@@ -256,7 +256,7 @@ module PuppetX::Firewall # rubocop:disable Style/ClassAndModuleChildren
       'wesp' => '141', 'rohc' => '142', 'ethernet' => '143', 'mptcp' => '262',
     }.freeze
 
-    PROTO_NAME_TO_NUMBER = lambda do
+    def self.build_proto_name_to_number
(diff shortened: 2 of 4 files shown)


jst-cyr:
@Thermikbomber : Thanks for the PR! I've approved the spec tests to run. The acceptance tests and the Puppet Core 9 tests cannot run on forks, but there is some feedback on the[ Puppet 8 run from the spec tests](https://github.com/puppetlabs/puppetlabs-firewall/actions/runs/33757619983/job/101086449603?pr=1316):

```
Error: Layout/RescueEnsureAlignment: `rescue` at 274, 4 is not aligned with `begin` at 259, 27.
Error: Performance/RegexpMatch: Use `match?` instead of `=~` when `MatchData` is not used. (https://github.com/fastruby/fast-ruby#regexp-vs-regexpmatch-vs-regexpmatch-vs-stringmatch-vs-string-vs-stringmatch-code-)
```

jst-cyr:
@Thermikbomber : Thanks for the fix, I reran the spec tests for 8 and it [found some more issues](https://github.com/puppetlabs/puppetlabs-firewall/actions/runs/33895294934/job/101102723653?pr=1316).

I'm also going to kick off the full acceptance test suite on a branch PR so we can get the full feedback for Puppet Core 9 and the OS platforms too.

jst-cyr:
Running CI for this PR here: https://github.com/puppetlabs/puppetlabs-firewall/pull/1317

Thermikbomber:
Alright @jst-cyr - I did a fix for te Robocop Checks after yor Msg. (Commit 1b0b0e4). Spec tests Puppet: ~> 8.0, Ruby Ver: 3.2) went through...

jst-cyr:
@Thermikbomber : Thanks for that fix! I brought that into the PR for running the CI.

**Acceptance Test Results:** https://github.com/puppetlabs/puppetlabs-firewall/actions/runs/33900696535?pr=1317

Full green CI for both unit and acceptance tests! 

This still requires an engineering review when they have the time for it, but at least we got the early feedback for the syntax issues on CI and validated the acceptance tests are all still good. We may need to rebase/retest in the future, but at least for now this is in a good place.

PR #1318, opened 2026-09-15
Fix support for multiple firewall mark matches

mattmix:
## Summary

This change adds support for multiple mark matches.

## Related ticket / issue

- Ticket / issue:

## Required checklist

- [ ] Linked the driving ticket or issue above where one exists (a `MODULES-*` ticket via the structured spec template, or a GitHub issue).
- [ ] Coverage gate is green (or the baseline is preserved -- coverage was not reduced).
- [ ] Platform matrix is green -- the PR's CI checks below cover it, or link a recent nightly Litmus run if the supported-OS acceptance tests do not run on this PR.
- [ ] `Co-Authored-By: Claude` trailer is present on the commits if this PR was AI-assisted.

## Review policy (outcome-based)

PRs are judged on outcomes, not line counts. Pick the lane that matches this change so
reviewers know what coverage to expect. See `CLAUDE.md` for the full policy.

- **Auto-merge** -- green CI required, no human review needed:
  dependabot bumps, modulesync / `pdk update` resync, `REFERENCE.md` regeneration, trivial doc fixes (typos, link fixes).
  Substantial doc changes -- e.g. a significant `README.md` rewrite -- still warrant a human review.
- **Single reviewer** -- green CI plus coverage preserved required:
  bug fixes, supported-platform additions, simple feature work.
- **Full review (2 or more reviewers)** -- green CI plus coverage plus senior sign-off:
  provider / type-API changes, security fixes, breaking changes.

This PR is (check one):

- [ ] Auto-merge
- [x] Single reviewer
- [ ] Full review (2+)

## Additional context

When an external service is managing some firewall rules that matches multiple marks, the module currently gives warnings like:

```
Warning: Provider returned data that does not match the Type Schema for `firewall[9003 cilium: any->pod redirect proxied traffic to host proxy]`
 Value type mismatch:
    * match_mark: ["! 0xe00/0xf00", "! 0x800/0xf00"] (expects an undef value or a match for Pattern[/^(?:!\s)?[a-fA-F0-9x]+(?:\/[a-fA-F0-9x]+)?$/], got Tuple)
```

PR #1319, opened 2026-09-17
(TESTS-1318) Fix support for multiple firewall mark matches

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-firewall/pull/1318)


jst-cyr:
**Acceptance Test Results:** https://github.com/puppetlabs/puppetlabs-firewall/actions/runs/35259811427?pr=1319

Almost all green. Only fails on AlmaLinux 8 and the failure matches the failure on the latest nightly against main, so it is not introduced by this PR.

Feedback to be provided on the PR.

